PR #2, opened 2021-08-30
Testable canonical application specification model

autodidaddict:
This will be a WIP PR for a while. When this PR is "done" and ready for review, we'll be able to have automated tests that certify the actions that need to be taken in a control loop. This is a "pure model" PR, and there will be a subsequent PR that adds `GenServer` and `NATS` and lattice support for actually taking the actions recommended by the model, as well as observing actual state.


lattice_controller_engine/lib/lattice_controller_engine/observed/lattice.ex
```diff
@@ -0,0 +1,474 @@
...
+  The root structure of an observed lattice. An observed lattice is essentially an
+  event sourced aggregate who state is determined by application of a stream of
+  lattice events
+  """
+  alias __MODULE__
+  alias LatticeControllerEngine.Observed.{Provider, Host, Instance, LinkDefinition}
+
+  # We need the keys to be there, even if they hold empty lists
+  @enforce_keys [:actors, :providers, :hosts, :linkdefs]
+  defstruct [:actors, :providers, :hosts, :linkdefs, :ocimap, :instance_tracking]
```
brooksmtownsend, line 14: Do we need at least empty maps for the ocimap and instance tracking too?

lattice_controller_engine/lib/lattice_controller_engine/observed/lattice.ex
```diff
@@ -0,0 +1,474 @@
...
+            "link_name" => link_name,
+            "annotations" => annotations
+          },
+          datacontenttype: "application/json",
+          source: source_host,
+          type: "com.wasmcloud.lattice.health_check_failed"
+        }
+      ) do
+    l
+  end
```
brooksmtownsend, line 104: This feels like something we should keep track of (health check status) but I'm fine if this is still a TODO. Just making sure we aren't doing nothing here on purpose

lattice_controller_engine/lib/lattice_controller_engine/observed/lattice.ex
```diff
@@ -0,0 +1,474 @@
...
+        source: _source_host,
+        datacontenttype: "application/json",
+        type: "com.wasmcloud.lattice.ocimap_del"
+      }) do
+    %Lattice{l | ocimap: Map.delete(l.ocimap, image_ref)}
+  end
+
+  def apply_event(l = %Lattice{}, _evt = %Cloudevents.Format.V_1_0.Event{}) do
+    l
+  end
```
brooksmtownsend, line 243: I was going to comment to add a warn here for an unexpected event, but I actually like this so we don't get super spammy logs as we add more events that the controller doesn't care about 👍🏻

lattice_controller_engine/lib/lattice_controller_engine/observed/lattice.ex
```diff
@@ -0,0 +1,474 @@
...
+          String.t()
+        ) :: [%{id: String.t(), instance_id: String.t(), host_id: String.t()}]
+  def running_instances(%Lattice{} = l, pk, spec_id) when is_binary(pk) do
+    if String.starts_with?(pk, "M") do
+      actors_in_appspec(l, spec_id)
+      |> Enum.map(fn %{actor_id: pk, instance_id: iid, host_id: hid} ->
+        %{id: pk, instance_id: iid, host_id: hid}
+      end)
+    else
+      providers_in_appspec(l, spec_id)
```
brooksmtownsend, line 407: Because this is the fallback, is there a possible case where the `spec_id` is invalid and not corresponding to an actor or a provider? if that's the case, you may need a `provider_in_appspec` definition that is a fallthrough for an invalid spec_id

lattice_controller_engine/lib/lattice_controller_engine/reconciler/app_spec.ex
```diff
@@ -0,0 +1,20 @@
...
+      Traits.reconcile_trait(desired, component, trait, actual)
+    end
+    |> List.flatten()
+  end
+
+  def matching_hosts(actual = %LatticeControllerEngine.Observed.Lattice{}, requirements = %{}) do
+    actual.hosts
+    |> Enum.filter(fn {_host_id, host} -> host.labels == requirements end)
+    |> Enum.map(fn {host_id, _host} -> host_id end)
+  end
```
brooksmtownsend, line 19: Are host labels required to be an exact match, or do we accept host labels that are a superset? e.g. if I need a host with `hostcore.arch == x86_64`, that should return a match even if that host has more labels

lattice_controller_engine/lib/lattice_controller_engine/reconciler/spreadscaler.ex
```diff
@@ -0,0 +1,126 @@
...
+        opts \\ %{}
+      ) do
+    for wtarget <- scaler.spread do
+      candidate_hosts = AppSpecReconciler.matching_hosts(actual, wtarget.requirements)
+
+      running_instances =
+        Lattice.running_instances(actual, component_pk, spec_name)
+        |> Enum.filter(fn %{host_id: hid} -> hid in candidate_hosts end)
+
+      desired_instances = trunc(wtarget.weight / 100 * scaler.replicas)
```
brooksmtownsend, line 33: PEMDAS scares me here, does elixir evaluate this left to right?

brooksmtownsend (changes requested):
Looks awesome so far. Just a couple questions and suggestions

lattice_controller_engine/lib/lattice_controller_engine/model/validator.ex
```diff
@@ -0,0 +1,97 @@
...
+            ]
+        }
+
+      true ->
+        result
+    end
+  end
+
+  defp validate_trait(%SpreadScaler{} = trait, result) do
+    weight_total = Enum.reduce(trait.spread, 0, fn x, acc -> acc + x.weight end)
```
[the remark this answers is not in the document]
autodidaddict: As a result I felt that a `reduce` here was a little more clear than a `map` to get the `weight` field followed by a `sum` call.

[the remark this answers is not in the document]
brooksmtownsend: Ah I see, you'd have to do some mapping in order to use the sum enumerable. Carry on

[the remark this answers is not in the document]
autodidaddict: After seeing it both ways I think it's actually slightly more readable as a map + sum combo.

lattice_controller_engine/lib/lattice_controller_engine/observed/host.ex
```diff
@@ -0,0 +1,15 @@
...
+  @enforce_keys [:id, :labels]
+  defstruct [:id, :labels, :last_seen]
+
+  @typedoc """
+  Represents a host observed through heartbeat events within a lattice
+  """
+  @type t :: %Host{
+          id: String.t(),
+          labels: Map.t(),
+          last_seen: Datetime.t()
```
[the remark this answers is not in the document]
autodidaddict: yes

lattice_controller_engine/lib/lattice_controller_engine/observed/lattice.ex
autodidaddict: Yes, we get those when we use the `Lattice.new` function

lattice_controller_engine/lib/lattice_controller_engine/observed/lattice.ex
autodidaddict: yes we'll ultimately want to do something meaningful with the health check pass fail. I've added this in #3

lattice_controller_engine/lib/lattice_controller_engine/observed/lattice.ex
autodidaddict: noted, good catch.

lattice_controller_engine/lib/lattice_controller_engine/reconciler/app_spec.ex
autodidaddict: D'OH.
Great catch here.

lattice_controller_engine/lib/lattice_controller_engine/reconciler/spreadscaler.ex
autodidaddict: `*` and `/` have the same precedence so this is left-to-right.  It's doing the percentage (/100) first, e.g. 50 / 100 * 6 == 0.5 * 6 == 3, and not 50 / 600 == 0.083333

autodidaddict: I had parens there but `mix format` actually removes them.

lattice_controller_engine/lib/lattice_controller_engine/model/decoder.ex
```diff
@@ -0,0 +1,225 @@
...
+  defp from_map(
+         %{
+           "name" => name,
+           "type" => @capability_component_type,
+           "properties" => %{
+             "contract" => contract,
+             "image" => image
+             # "link_name" => link_name
+           }
+           # "traits" => traits
```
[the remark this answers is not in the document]
autodidaddict: Removed commented fields.

lattice_controller_engine/lib/lattice_controller_engine/reconciler/spreadscaler.ex
brooksmtownsend: Cool, just making sure. This is one of those bugs that could take hours and would make you want to smash your machine, so I figured I'd just double check

lattice_controller_engine/test/reconciler/scaler_test.exs
```diff
@@ -78,7 +78,8 @@ defmodule LatticeControllerEngineTest.Reconciler.ScalerTest do
             "N00001" => %{
               id: "N00001",
               labels: %{
-                "style" => "cool"
+                "style" => "cool",
+                "extraneous" => "yup"
```
brooksmtownsend, line 82: LOL

brooksmtownsend (approved):

PR #7, opened 2021-11-27
mix test in the wadm repository

janitha09:
Wadm.Observed.Instance.__struct__/1 is undefined, cannot expand struct Wadm.Observed.Instance. Make sure the struct name is correct. If the struct name exists and is correct but it still cannot be found, you likely have cyclic module usage in your code
https://github.com/search?q=org%3AwasmCloud+Wadm.Observed&type=code

wadm.Observed.* is not defined as a module. Something close exists in LatticeObserver.Observed.*. However :ocimap is missing from the struct
https://github.com/wasmCloud/lattice-observer/blob/73e937dc81438b9c5bd720402a8ea4a3994b0a17/lib/lattice_observer/observed/lattice.ex#L16

The changes here make hacky changes the let the tests run. 13 out of the 20 tests fail.

Am I doin the correct thing?


wadm/mix.exs
```diff
@@ -24,7 +24,7 @@ defmodule Wadm.MixProject do
       {:yaml_elixir, "~> 2.8"},
       {:cloudevents, "~> 0.4.0"},
       {:uuid, "~> 1.1"},
-      {:lattice_observer, git: "https://github.com/wasmcloud/lattice-observer"}
+      {:lattice_observer, git: "https://github.com/janitha09/lattice-observer"}
```
autodidaddict, line 27: Per previous discussion we don't need this one-off, the fix is to make wadm use the `refmap` field, not the `ocimap` field

autodidaddict (changes requested):
I think if we refactor wadm to use the `refmap` field instead of the `ocimap` field, things should be good. Otherwise your changes to use the lattice observer structs look good.

Also, please click the "details" button on the failing DCO check for instructions on how to fix that.

autodidaddict:
It's also entirely possible that recent changes to the lattice observer haven't been carried back into wadm and so some of the data structures may be shaped differently. You can see the data structure changes if you look at the recent commit history for lattice observer.

autodidaddict:
Closing this PR because I incorporated your changes in #8

PR #8, opened 2021-11-28
Fixes remaining tests

autodidaddict:
@janitha09 - I messed up while trying to submit a pull request to your branch so I ended up here - your original commits are still in there.

tl;dr - I took what you provided and then fixed the remaining tests.

this should replace PR #7


autodidaddict:
@janitha09 please fix the DCO check and then we should be ready to merge.

autodidaddict:
Upon coming to my senses, I realized that all of the tests in the `observed` folder are just duplicates of the tests that are inside the lattice observer. now that these are gone, it should make this codebase a bit more resilient against internal data structure changes to the observed lattice.

PR #10, opened 2021-12-15
[WIP] attempt at hydrating the state i.e. lattice maintained by Wadm

janitha09:
@autodidaddict @brooksmtownsend 

I think handle_info is missing (if wadm is listening)

probably something along these lines from OTP
```
  @impl true
  def handle_info(
        {:msg, %{body: body, topic: topic}},
        state
      ) do
    Logger.debug("StateMonitor handle info #{topic}")

    state =
      if String.starts_with?(topic, "wasmbus.evt.") do
        handle_event(state, body)
      end

    {:noreply, state}
  end
```
In the IT tests that I am trying to run the state needs to be hydrated when OTP is and nats are already up and running.


autodidaddict:
This is pretty close. I think the only thing missing is that somewhere in the `wadm` startup sequence, it should be calling the `LatticeObserver.NatsObserver` init with the supervised connection that wadm uses. I'll get another PR ready that puts this in place as an illustration.

Also it looks like there's extraneous history in this PR (I think maybe your fork hasn't been updated to `main`?) I'm not sure, my github-fu is pretty weak.

brooksmtownsend:
> Also it looks like there's extraneous history in this PR (I think maybe your fork hasn't been updated to main?) I'm not sure, my github-fu is pretty weak.

Yeah this just looks like the fork needs a rebase on main

janitha09:
I agree that the supervisor should be in the app startup, that seems an incomplete solution to me, we will have a timing dependency i.e. wadm must be up before otp is up. What do we do on a restart or crash of the application. In integrity scan/poll makes sense. I am looking at whether I can implement something that can query the :state_monitor genserver. Which implies that I barely understand (do not have a strong enough mental model) what the lattice observer does i.e. what genservers from otp we are observing via the cloud events on NATS

janitha09:
Will address comments through different PR

PR #12, opened 2021-12-27
Adds integration with lattice observer library + new supervision hierarchies

autodidaddict:
To play with this:
```
Interactive Elixir (1.12.1) - press Ctrl+C to exit (type h() ENTER for help)

11:22:04.389 [debug] connecting to %{host: '127.0.0.1', port: 4222}
iex(1)> spec = Wadm.Model.AppSpec.new("test", "0.1.0", "")
%Wadm.Model.AppSpec{
  components: [],
  description: "",
  name: "test",
  version: "0.1.0"
}
iex(2)> {:ok, pid} = Wadm.Deployments.Supervisor.start_deployment_monitor(spec, "default")

11:23:38.516 [debug] Starting event processor for deployment test v0.1.0
 
11:23:38.516 [debug] Starting new lattice observer for prefix 'default'
{:ok, #PID<0.305.0>}
iex(3)> 
11:23:44.636 [debug] Processing event com.wasmcloud.lattice.host_started
```

This will give you a new `DeploymentMonitor` process managing the state of a deployed appspec. Each time you fire up a deployment monitor, if one is not already running, it creates a `LatticeObserver.NatsObserver` (from the lattice-observer library). This will then set up message forwarding from the lattice observer to the deployment monitor where you can see the scaffolding for performing a reconciliation and then (not yet written) emitting those tasks to a distributed work dispatch queue.

There are plenty of areas for improvement here @janitha09 - such as migrating the explicit send paths to a chain of `GenStage` so we can have flow control, but this should provide a decent set of scaffolding to give you an idea of where my head was at in terms of implementation for wadm.


wadm/lib/wadm/observer/observer.ex
```diff
@@ -0,0 +1,37 @@
...
+  It's important to note that this module isn't stateful nor is it a genserver.
+  There will be 1 monitor process per lattice prefix, so the more of those that exist,
+  the more lattice traffic will pass through this 'state_changed' function.
+  """
+  use LatticeObserver.Observer
+  import Wadm.Deployments.DeploymentMonitor, only: [is_reconciliation_ready?: 1]
+  require Logger
+
+  @impl true
+  def state_changed(old_state, new_state, event, lattice_prefix) do
```
thomastaylor312, line 16: Generally speaking if we are making this declarative, we shouldn't pass old state. All that matters is how the new state differs from the current state. Based on my experience from having reconciliation loops in k8s, passing old_state into a handler leads to bad coding practices where people try to carry all sorts of weird state in the struct/event data that should just be available from the system

thomastaylor312 (approved):
Looking pretty good. Just one minor, non-blocking nit on how we are handling state changes

wadm/lib/wadm/observer/observer.ex
autodidaddict: yeah the function name `state_changed` is misleading. It is actually invoked every time an event is received regardless of whether it made a difference in state. The reason I passed both was to allow the watchers/consumers to decide if the state change made in the processing of the event was worth acting upon or not.

autodidaddict: The more I re-read your comment the more I think I should change the latticeobserver behaviour so that the callback is called `event_received` (or something similar) and it only contains the new/current state, event, and prefix and not the old state.

autodidaddict:
going to go ahead and merge this and we can address more of the other comments in forthcoming PRs.

PR #13, opened 2021-12-29
Initial commit of high availability clustering support

autodidaddict:
As of this PR, a given _cluster_ of nodes in a "wadm cluster" will now evenly share process instances for deployment managers (1 per lattice+spec+version) and for lattice observers (NATS, 1 per lattice). If a node goes down, then all of these Horde-supervised processes will be immediately restarted on other, suitable nodes. After restarting, they will attempt to recover their state (if applicable) from redis.

I'll move **lattice-observer** back to a git reference once the active PR there is merged.

p.s. 🤘 _For the Horde!_ 🤘
image


wadm/lib/supervisor.ex
```diff
@@ -8,6 +8,12 @@ defmodule Wadm.Supervisor do
   def init(_opts) do
     config = Vapor.load!(Wadm.ConfigPlan)
 
+    topologies = [
+      wadmcluster: [
+        strategy: Cluster.Strategy.Gossip
```
thomastaylor312, line 13: Is this using the multicast you told me about before? If so, will we do a follow up to use something NATS based?

wadm/lib/wadm/deployments/deployment_monitor.ex
```diff
@@ -1,18 +1,41 @@
...
+  Process Drift
+  When a node in a wadm cluster shuts down (intentionally or otherwise), this process
+  will be re-started on a new target node _without saving state_. An instant later, that
+  same process will try and reconstitute its state from cache.
   """
   use GenServer
   require Logger
+  alias LatticeObserver.Observed.Lattice
+
+  @decay_timer 39_000
```
thomastaylor312, line 20: Nit: Is this ms? We should probably change the constant to indicate units

wadm/lib/wadm/observer/cache.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Wadm.Observer.Cache do
+  require Logger
+
+  # return empty for now
+  # TODO: load from a real cache
```
thomastaylor312, line 5: Looks like you are loading from cache now, so this can be deleted

wadm/mix.exs
```diff
@@ -24,7 +24,12 @@ defmodule Wadm.MixProject do
       {:yaml_elixir, "~> 2.8"},
       {:cloudevents, "~> 0.4.0"},
       {:uuid, "~> 1.1"},
-      {:lattice_observer, git: "https://github.com/wasmcloud/lattice-observer"},
+      {:libcluster, "~> 3.3"},
+      {:horde, "~> 0.8.5"},
+      {:gnat, "~> 1.4"},
+      {:redix, "~> 1.1"},
+      {:lattice_observer, path: "../../lattice-observer"},
+      # {:lattice_observer, git: "https://github.com/wasmcloud/lattice-observer"},
```
thomastaylor312, line 32: Does this need to be uncommented before merge?

thomastaylor312 (approved):

wadm/lib/supervisor.ex
autodidaddict: Yeah. There's a stack of already-available cluster strategies (including multiple DNS-based and a k8s-based). We can write our own NATS-based one if we see the need to expand on what comes "stock"

wadm/mix.exs
autodidaddict: I'll get switched back to git once the lattice-observer pending PR is in main

autodidaddict: But that can happen in another PR

PR #14, opened 2021-12-31
Initial commit of wadm API scaffolding

autodidaddict:
This is, for the moment, a WIP. Once we like the API and shapes, then I can go through and make all the operations real.

### DIFF
--- a/wadm/lib/wadm/api/api_server.ex
+++ b/wadm/lib/wadm/api/api_server.ex
@@ -1,0 +1,174 @@
+defmodule Wadm.Api.ApiServer do
+  @moduledoc """
+  Exposes the API for interacting with wadm over the NATS subject wadm.api.>
+  The following is a list of the operations available in this API (minus the prefix):
+
+  * `apps.get` - Retrieves a summary of the application models stored by this wadm cluster. This will include status information
+  * `app.put` - Adds a new application model
+  * `app.del` - Deletes an application model
+  * `app.deploy` - Deploys an application to a given lattice
+  * `app.undeploy` - Stops managing an application in the given lattice. Non-destructive.
+  * `app.details` - Retrieves a full set of details about an application, including the model
+  * `app.history` - Retrieves an audit history for a particular application
+  """
+  require Logger
+  use Gnat.Server
+
+  alias Wadm.Model.{Validator, AppSpec}
+
+  def request(%{topic: topic, body: body}) do
+    topic
+    |> String.split(".")
+    # wadm
+    |> List.delete_at(0)
+    # api
+    |> List.delete_at(0)
+    |> List.to_tuple()
+    |> handle_request(body)
+  end
+
+  # TODO just mocking this data up for now.
+  # will make it real once we like the shapes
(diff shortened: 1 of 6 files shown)


wadm/lib/wadm/api/api_server.ex
```diff
@@ -0,0 +1,174 @@
...
+  end
+
+  defp handle_request({"app", "del"}, body) do
+    with {:ok, _request} <- Jason.decode(body) do
+      # TODO - remove from store
+      # %{
+      # "name" => "name",
+      # "versions" => []
+      # }
+      # if the list of versions is empty, _all_ versions will be removed
```
thomastaylor312, line 83: Are we going to support running multiple versions of the same app model?

wadm/lib/wadm/api/api_server.ex
```diff
@@ -0,0 +1,174 @@
...
+          ]
+        }
+      ]
+      |> Jason.encode!()
+
+    {:reply, response}
+  end
+
+  defp handle_request({"app", "put"}, body) do
+    # input here is a JSON version of the OAM model
```
thomastaylor312, line 54: Just to double check, is OAM what we want to use? Is that going to be too much overhead for what we need wadm to do?

thomastaylor312:
Initial structure looks pretty good so far. No major concerns with the data shape right now

wadm/test/model/yaml_json_test.exs
```diff
@@ -1,18 +1,26 @@
-defmodule WadmTest.Model.YamlTest do
+defmodule WadmTest.Model.YamlJsonTest do
```
brooksmtownsend, line 1: integration tests from an initial commit? 😍

wadm/lib/wadm/api/api_server.ex
```diff
@@ -0,0 +1,174 @@
+defmodule Wadm.Api.ApiServer do
+  @moduledoc """
+  Exposes the API for interacting with wadm over the NATS subject wadm.api.>
+  The following is a list of the operations available in this API (minus the prefix):
+
+  * `apps.get` - Retrieves a summary of the application models stored by this wadm cluster. This will include status information
+  * `app.put` - Adds a new application model
+  * `app.del` - Deletes an application model
+  * `app.deploy` - Deploys an application to a given lattice
+  * `app.undeploy` - Stops managing an application in the given lattice. Non-destructive.
```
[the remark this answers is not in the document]
autodidaddict: yeah I'm thinking of adding a `destructive` option to the `undeploy` command, that will, if set, terminate all the resources that were being managed by that deployment.

wadm/lib/wadm/api/api_server.ex
autodidaddict: Only a single version of a particular spec will be deployed in any given lattice at one time.

thomastaylor312: Ok, so I read this comment as us supporting multiple versions. We may want to clarify

PR #16, opened 2022-01-12
Adds support for mnesiac

janitha09:


wadm/lib/wadm/deployments/deployment_monitor.ex
```diff
@@ -19,6 +19,8 @@ defmodule Wadm.Deployments.DeploymentMonitor do
 
   @decay_timer_ms 39_000
 
+  @configuredcache [Wadm.Observer.RedisCache]
```
janitha09, line 22: replace with Wadm.Observer.MnesiacCache one could also send to both redis and mnesia for comparison.

janitha09:
@autodidaddict

brooksmtownsend:
@janitha09 after the community meeting, were there more changes that you are making to this PR or is it ready-to-go?

janitha09:
@brooksmtownsend There are more changes, particularly around the interaction between Horde libcluster and mnesiac that Kevin alluded to. While the supervisor does start mnesiac the use of it is optional at this time. It is safe to merge. It does not prevent the use of Redis. This is a feature that saves the user from having to maintain Redis long term it is something we want to get working and be the default

autodidaddict:
Going to close this until such time as amnesiac works cleanly by default with dynamically added nodes.

PR #18, opened 2022-06-10
Preliminary implementation of model store and retrieve

autodidaddict:
Initial implementation of CRUD functionality on app specs. Also implements deploy/undeploy functionality on the agent process, though the agent process currently does nothing.

### DIFF
--- a/wadm/lib/supervisor.ex
+++ b/wadm/lib/supervisor.ex
@@ -27,7 +27,7 @@ defmodule Wadm.Supervisor do
         id: :api_connection_supervisor
       ),
       Supervisor.child_spec(
-        {Redix, host: System.get_env("REDIS_HOST") || "127.0.0.1", name: :model_store},
+        {Redix, host: config.redis.host, name: :model_store},
         id: :model_store
       ),
       Supervisor.child_spec(
--- a/wadm/lib/supervisor.ex
+++ b/wadm/lib/supervisor.ex
@@ -27,6 +27,10 @@ defmodule Wadm.Supervisor do
         id: :api_connection_supervisor
       ),
       Supervisor.child_spec(
+        {Redix, host: System.get_env("REDIS_HOST") || "127.0.0.1", name: :model_store},
+        id: :model_store
+      ),
+      Supervisor.child_spec(
         {Gnat.ConsumerSupervisor,
          %{
            connection_name: :api_nats,
--- a/wadm/lib/wadm/api/api_server.ex
+++ b/wadm/lib/wadm/api/api_server.ex
@@ -27,6 +27,7 @@ defmodule Wadm.Api.ApiServer do
         {:reply, success_result(%{})}
       else
         Logger.debug("Stopping deployment monitor: #{model_name}")
+
         case Horde.DynamicSupervisor.terminate_child(Wadm.HordeSupervisor, monitor) do
           :ok ->
             {:reply, success_result(%{})}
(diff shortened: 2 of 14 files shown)


wadm/lib/wadm/model/store.ex
```diff
@@ -0,0 +1,222 @@
...
+            vetted: Wadm.Model.AppSpec.t()
+          }
+
+    @derive Jason.Encoder
+    defstruct [:raw, :vetted]
+  end
+
+  @type model_map :: Map.t()
+
+  def lattice_models(redis, lattice_id) when is_binary(lattice_id) do
```
connorsmith256, line 34: This could use a doc indicating it returns the most recent version for each model

connorsmith256 (approved):

wadm/lib/supervisor.ex
```diff
@@ -26,6 +26,10 @@ defmodule Wadm.Supervisor do
         {Gnat.ConnectionSupervisor, Wadm.Api.Connection.settings_from_config(config)},
         id: :api_connection_supervisor
       ),
+      Supervisor.child_spec(
+        {Redix, host: System.get_env("REDIS_HOST") || "127.0.0.1", name: :model_store},
```
thomastaylor312, line 30: Nit: Should probably be `WADM_REDIS_HOST`

wadm/lib/wadm/model/store.ex
```diff
@@ -0,0 +1,222 @@
...
+      true
+    else
+      _ ->
+        false
+    end
+  end
+
+  defp stamp_to_rfc_3339(stamp) when is_integer(stamp) do
+    with {:ok, dt} <- DateTime.from_unix(stamp, :millisecond) do
+      dt |> DateTime.to_iso8601()
```
thomastaylor312, line 207: So this goes to iso8601 and not rfc3339. They are basically the same, but as far as I understand the `T` separator is the main difference. I don't care which one we use, but we should make sure it is the right format

thomastaylor312 (changes requested):
One super minor change and a few nits

wadm/lib/wadm/api/api_server.ex
```diff
@@ -27,148 +19,249 @@ defmodule Wadm.Api.ApiServer do
...
     end
   end
 
-  defp handle_request({"app", "details"}, body) do
-    with {:ok, _request} <- Jason.decode(body) do
-      # TODO - load from store
-      # this should return the OAM model that was sent in a "push" and NOT
-      # the internal %AppSpec{} struct
+  # Queries the list of versions available for a given model in the given lattice
+  # Versions will be returned in order of newest-first
```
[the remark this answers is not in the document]
autodidaddict: this comes from the timestamp of the revision being the score in the scored set. when we do `REV` on the output, we get it in descending order.

wadm/lib/wadm/model/store.ex
```diff
@@ -0,0 +1,222 @@
...
+        {:error, e}
+    end
+  end
+
+  @spec delete_model_version(Redix.connection(), String.t(), String.t(), String.t()) ::
+          :ok | {:error, String.t()}
+  @doc """
+  This function removes a model version from the persistent store.
+  """
+  def delete_model_version(redis, model_name, version, lattice_id)
```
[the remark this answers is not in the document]
autodidaddict: need to drop the model name. Good catch

PR #19, opened 2022-07-06
Initial implementation of deploy/undeploy functionality

autodidaddict:
NOTE that this is currently using a naive state model and a somewhat naive backoff timer during reconciliation. I fully expect all of this functionality to improve over time but in the interest of giving people something to iterate on, this basic implementation is contained in this PR.

### DIFF
--- a/wadm/lib/wadm/deployments/deployment_monitor.ex
+++ b/wadm/lib/wadm/deployments/deployment_monitor.ex
@@ -160,7 +160,19 @@ defmodule Wadm.Deployments.DeploymentMonitor do
   end
 
   defp publish_lattice_control_command({{topic, cmd}, lattice_id}) do
-    _ = Gnat.request(String.to_atom(lattice_id), topic, cmd)
+    case Gnat.request(String.to_atom(lattice_id), topic, cmd) do
+      {:ok, res} ->
+        case Jason.decode(res) do
+          {:ok, %{"accepted" => false, "error" => err}} ->
+            Logger.error("Lattice control interface rejected request: #{err}")
+
+          _ ->
+            Logger.debug("Successfully performed lattice control interface request")
+        end
+
+      {:error, _} ->
+        Logger.error("Timeout occurred attempting to make lattice control interface request")
+    end
   end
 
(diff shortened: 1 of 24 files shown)


wadm/lib/wadm/deployments/deployment_monitor.ex
```diff
@@ -43,24 +59,110 @@ defmodule Wadm.Deployments.DeploymentMonitor do
...
+    # even if we had a failure in the command list, back off so that we don't attempt
+    # to re-reconcile for another 30 seconds
+    Process.send_after(self(), :remove_backoff, 45_000)
+
+    {length(cmds), length(errors)}
+  end
+
+  defp publish_lattice_control_command({{topic, cmd}, lattice_id}) do
+    _ = Gnat.request(String.to_atom(lattice_id), topic, cmd)
+  end
```
brooksmtownsend, line 164: This would be a good place to error log if a lattice control request fails to receive a reply, or even when a request is malformed

brooksmtownsend (changes requested):
two smaller nits, but the real request i have is for a "How to run wadm" section in the README, even if it just involves an elixir prerequisite and `mix` for now

wadm/lib/wadm/deployments/deployment_monitor.ex
```diff
@@ -43,24 +59,110 @@ defmodule Wadm.Deployments.DeploymentMonitor do
...
+    Logger.debug("Reconciliation found #{length(skips)} no-ops and #{length(cmds)} commands")
+
+    cmds
+    |> Enum.map(fn cmd ->
+      {Wadm.Reconciler.Command.to_lattice_control_command(spec, lattice.id, cmd), lattice.id}
+    end)
+    |> Enum.each(&publish_lattice_control_command/1)
+
+    # even if we had a failure in the command list, back off so that we don't attempt
+    # to re-reconcile for another 30 seconds
```
connorsmith256, line 156: Nit: mismatch between 30 and 45

connorsmith256 (approved):

wadm/lib/wadm/deployments/deployment_monitor.ex
connorsmith256, line 156: Nit: mismatch between 30 and 45 seconds

PR #27, opened 2022-08-15
Corrected argument orders for events

brooksmtownsend:
This PR corrects a few event arguments that were ordered incorrectly, or in the `version` case sending the entire model instead of just the version.

Since #26 is now merged into main, this version may need to bump to `0.2.0`. Thoughts @autodidaddict ?


connorsmith256 (approved):
:reverse:

PR #28, opened 2022-08-16
Allow linkdefs to put during reconciliation loop

brooksmtownsend:
This PR:
- Adds more instructions into the README for how to run wadm
- Wraps NATS calls in the safe wrapper like host_core to ensure no crashes when connection is lost
- Reduces the startup time by a few seconds, initiating the host ping quicker
- Refactors a few functions, notably allowing a reconcile to perform specific actions or calculate what to do
- Allows linkdefs to be put during the reconciliation loop

The last point is a big one, since it was a deliberate decision to only issue lattice control commands when wadm is not reconciling to avoid thrashing. I still think it's a good idea to prevent starting and stopping actors/providers because those are _not_ idempotent and change the state of the lattice in a way that can thrash. However, link definitions are idempotent and can be put multiple times without any side effects.

In the future, if wadm takes on the responsibility of deleting linkdefs on an undeploy or something similar, then we'll want to revisit this.

### DIFF
--- a/wadm/lib/wadm/deployments/deployment_monitor.ex
+++ b/wadm/lib/wadm/deployments/deployment_monitor.ex
@@ -84,22 +84,25 @@ defmodule Wadm.Deployments.DeploymentMonitor do
     # Make sure that there's a lattice supervisor running
     {:ok, _pid} = Wadm.LatticeSupervisor.start_lattice_supervisor(state.lattice_id)
 
-    Process.send_after(self(), :host_ping, 5_000)
+    Process.send_after(self(), :host_ping, 1_000)
     {:noreply, state}
   end
 
   @impl true
   def handle_info(:host_ping, state) do
+    # We pub instead of requesting here as we don't actually need to use the host
+    # inventories here, we create lattice state based on the heartbeats that are published
     if Wadm.Nats.safe_pub(
          String.to_atom(state.lattice_id),
          "wasmbus.ctl.#{state.lattice_id}.ping.hosts",
          ""
        ) != :ok do
       Logger.warning(
-        "Timed out on host ping, first reconcilation pass may fail without proper host information"
+        "Failed to pings, first reconcilation pass may fail without proper host information"
       )
     end
 
+    # Give hosts across a lattice 5 seconds to send their inventories
     Process.send_after(self(), :reconcile_initializing, 5_000)
     {:noreply, state}
   end
@@ -213,9 +216,6 @@ defmodule Wadm.Deployments.DeploymentMonitor do
               put_cmds ->
                 do_reconcile(state.spec, lattice, {put_cmds, [], []})
             end
-
-          _ ->
-            :ok
         end
 
         # Regardless if we took action or not, we will check at the end of the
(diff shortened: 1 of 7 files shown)


wadm/lib/wadm/deployments/deployment_monitor.ex
```diff
@@ -84,20 +84,25 @@ defmodule Wadm.Deployments.DeploymentMonitor do
...
-    if Gnat.pub(
+    # We pub instead of requesting here as we don't actually need to use the host
+    # inventories here, we create lattice state based on the heartbeats that are published
+    if Wadm.Nats.safe_pub(
          String.to_atom(state.lattice_id),
          "wasmbus.ctl.#{state.lattice_id}.ping.hosts",
          ""
        ) != :ok do
-      Logger.warn("Could not initiate host ping, first reconcilation pass may fail")
+      Logger.warning(
```
connorsmith256, line 100: I just realized there's both `warn` and `warning` macros :eye-twitch:

Apparently `warn` is deprecated? Good to know, but sad the more verbose option is preferred...

wadm/lib/wadm/deployments/deployment_monitor.ex
```diff
@@ -195,9 +200,26 @@ defmodule Wadm.Deployments.DeploymentMonitor do
...
+      # as they are idempotent and can be applied multiple times (in the worst case)
+      # without side effects.
+      # Other events have no effect but will trigger a check after cooling off
       :reconciling ->
+        case actions_to_reconcile(state.spec, lattice) do
+          {cmds, _skip, _err} ->
+            case Enum.filter(cmds, fn %Wadm.Reconciler.Command{cmd: cmd} ->
+                   cmd == :put_linkdef
+                 end) do
+              [] ->
```
connorsmith256, line 213: Does `do_reconcile` work with an empty list of commands? If so removing this `case` would make this code more readable

wadm/lib/wadm/nats.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Wadm.Nats do
+  @moduledoc false
+  require Logger
+
+  # Heavily inspired from the flawless https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/nats.ex
+  def safe_pub(process_name, topic, msg) do
+    if Process.whereis(process_name) != nil do
+      Gnat.pub(process_name, topic, msg)
+    else
+      Logger.warn("Publication on #{topic} aborted - connection #{process_name} is down",
```
connorsmith256, line 10: Should this be `warning`?

brooksmtownsend: yeah good catch, I'm used to warn but I noticed the deprecation on the other one :(

wadm/lib/wadm/nats.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Wadm.Nats do
+  @moduledoc false
+  require Logger
+
+  # Heavily inspired from the flawless https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/nats.ex
```
autodidaddict, line 5: :rofl:

autodidaddict (approved):

connorsmith256 (approved):

PR #33, opened 2022-08-25
Fixed a few logic errors in config

brooksmtownsend:
Previously the port would be interpreted as a string which fails to connect. Additionally, I simplified the secret storage to just `seed` instead of `nkey_seed`

edit: Also I noticed that NATS wasn't running for the tests so they were all technically failing. Added that step

### DIFF
--- a/wadm/README.md
+++ b/wadm/README.md
@@ -51,7 +51,7 @@ The following environment variables are used for wadm's vault integration:
 | `WADM_VAULT_ADDR` | `http://127.0.0.1:8200` | The address of the vault instance |
 | `WADM_VAULT_PATH_TEMPLATE` | `(Empty)` | A template used to define the vault path for lattice user credentials |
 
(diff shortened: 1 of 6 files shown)


wadm/README.md
```diff
@@ -35,6 +35,51 @@ mix do deps.get, compile
...
+
+The following environment variables are used for wadm's vault integration:
+
+| Variable | Default Value | Description |
+| --- | --- | --- |
+| `WADM_VAULT_TOKEN` | `(Empty)` | A token used to authenticate against a vault instance |
+| `WADM_VAULT_ADDR` | `http://127.0.0.1:8200` | The address of the vault instance |
+| `WADM_VAULT_PATH_TEMPLATE` | `(Empty)` | A template used to define the vault path for lattice user credentials |
+
+The credentials stored within a vault secret are pushed as a JSON payload containing the fields `jwt` and `nkey_seed`. When stored properly, you can use the `vault` command line to inspect the credentials as shown below:
```
autodidaddict, line 54: change nkey_seed to seed here.

autodidaddict (changes requested):
1 nit

autodidaddict (approved):

PR #34, opened 2022-08-29
Default to info level logs in prod

brooksmtownsend:
This is largely due to the fact that the Gnat library logs connection information at the debug level (e.g. jwt + seed used for connecting)


connorsmith256 (approved):
Does `logger` allow you to configure on a per-library basis, so we could set just Gnat to info?

brooksmtownsend:
@connorsmith256 figured it out!

connorsmith256 (approved):
Noice!

PR #35, opened 2022-08-29
feat(*): Add better support for non-conditional spreadscaler

thomastaylor312:
This PR also serves as a proposal. As things currently stand, a user
has to specify a full set of requirements and name in order to deploy
multiple replicas:

```yaml
- type: spreadscaler
  properties:
    replicas: 1
    spread:
      - name: customersclinicapp
        requirements:
          app: petclinic
```

However, if a user wanted to just run a certain number of replicas, without
setting up tagged hosts (e.g. when running `wash up`), they would get an
error about having an invalid trait if the `spread` key wasn't included. This
PR adds support for setting only the `replicas` field when you don't care
about requirements:

```yaml
- type: spreadscaler
  properties:
    replicas: 1
```

It also allows for `nil` or empty requirements fields for when you might want
to specify weight for all hosts and then run some on another host:

```yaml
- type: spreadscaler
  properties:
    replicas: 1
    spread:
      - name: allhosts
        weight: 66
      - name: customersclinicapp
        weight: 33
        requirements:
          app: petclinic
```

### DIFF
--- a/wadm/test/fixtures/mixed_spreads.yaml
+++ b/wadm/test/fixtures/mixed_spreads.yaml
@@ -1,0 +1,36 @@
+apiVersion: core.oam.dev/v1beta1
+kind: Application
+metadata:
+  name: my-example-app
+  annotations:
+    version: v0.0.1
+    description: "This is my app"
+spec:
+  components:
+    - name: userinfo
+      type: actor
+      properties:
+        image: wasmcloud.azurecr.io/fake:1
+      traits:
+        - type: spreadscaler
+          properties:
+            replicas: 4
+            spread:
+              - name: allhosts
+                weight: 67
+              - name: westcoast
+                requirements:
+                  zone: us-west-1
+                weight: 33
+        - type: linkdef
+          properties:
+            target: webcap
+            values:
+              port: 8080
+
+    - name: webcap
+      type: capability
+      properties:
+        contract: wasmcloud:httpserver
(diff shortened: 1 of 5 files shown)


wadm/test/fixtures/mixed_spreads.yaml
```diff
@@ -0,0 +1,36 @@
...
+          properties:
+            target: webcap
+            values:
+              port: 8080
+
+    - name: webcap
+      type: capability
+      properties:
+        contract: wasmcloud:httpserver
+        image: wasmcloud.azurecr.io/httpserver:0.13.1
```
brooksmtownsend, line 35: What are we, in the stone ages here?

brooksmtownsend (commented):
Code looks good, and I like the simplification of the YAML. Wanted to let @autodidaddict leave thoughts before the ✅

wadm/test/fixtures/mixed_spreads.yaml
thomastaylor312: Whoops. Better go update the examples code I copied this from

thomastaylor312: Actually this doesn't matter. It is a fixture 🙃

brooksmtownsend: Yeah I was just jesting :)

autodidaddict (approved):

PR #60, opened 2023-03-05
feat(*): Completes spike of HA stream architecture

### LINKED ISSUE #42: Spike: Test out HA streaming architecture
See for context: https://github.com/wasmCloud/wadm/issues/40#issuecomment-1440686958

The definition of done is that the basic scaffolding is in place and working so we can make a decision whether or not to move forward with it. This scaffolding includes:
- Creating the actual wadm binary with the following functionality
  - No configuration needed for this first draft, we can add flags later
  - Create the needed streams for events and commands
  - Have spawned functions that
    - Listen on the consumer for events
    - Listen on the consumer for commands

thomastaylor312:
This adds some code for handling the event streams and a basic setup of the wadm binary. Much of the code in the binary will likely need to be refactored into library code, but it is good enough to prove that things work. I might do another follow on PR to this to maybe clean up a few more things.

Sorry for how giant this is, it was what was needed to get things working

Closes #42


bin/main.rs
```diff
@@ -1,3 +1,372 @@
...
+        }
+    }
+    Ok(())
+}
+
+// For testing only, to send a fake command in response to an event
+async fn send_fake_command(context: &Context, host_id: &str, event: &Event) -> anyhow::Result<()> {
+    use wadm::nats_utils::ensure_send;
+    let command: Command = match event {
+        Event::ActorStarted(actor) => {
```
autodidaddict, line 187: I know this is test, so I'm assuming this is just mock data... but wadm wouldn't emit a start actor command in response to an actor started event.

bin/main.rs
```diff
@@ -1,3 +1,372 @@
...
+            res?
+        }
+    }
+    Ok(())
+}
+
+// NOTE: These functions will probably need to be replaced with something in the library code, but
+// they're good enough for testing
+async fn handle_events(stream: Stream, context: Context, host_id: String) -> anyhow::Result<()> {
+    let mut consumer = EventConsumer::new(stream, "wasmbus.evt.default")
```
autodidaddict, line 150: assuming this hard coded topic is just because this is a test/experiment?

autodidaddict:
Given that this is experimental to prove that certain aspects of the design work, this all LGTM. Just take care to make sure that all the hardcoded bits (e.g. using `default` as the lattice name) become configurable bits as this gets migrated

autodidaddict (approved):

bin/main.rs
thomastaylor312: Yep, totally something to demo it works and so I knew what would be coming out the other end!

bin/main.rs
thomastaylor312: Yes, 100% hard coded. If you take a look at #61 you'll see that I made something to support multiple.

src/consumers/commands.rs
```diff
@@ -0,0 +1,123 @@
+//! A module for creating and consuming a stream of events from a wasmcloud lattice
```
brooksmtownsend, line 1: nit: copy paste comment

brooksmtownsend (approved):
Love the tests, architecture seems solid. I'll leave @autodidaddict to some details of HA, I left some nits

thomastaylor312:
Thanks for the ideas @brooksmtownsend! I'll wrap those nits up in #61

PR #61, opened 2023-03-06
feat(*): Cleaner task running

thomastaylor312:
Puts together something for actually doing work and managing several loops. This makes it easier to encapsulate the actual work we have to do down the line

### DIFF
--- a/src/consumers/manager.rs
+++ b/src/consumers/manager.rs
@@ -1,0 +1,168 @@
+use std::fmt::Debug;
+use std::{collections::HashMap, marker::PhantomData, sync::Arc};
+
+use async_nats::jetstream::stream::Stream as NatsStream;
+use futures::{Stream, StreamExt};
+use tokio::{
+    sync::{RwLock, Semaphore},
+    task::JoinHandle,
+};
+use tracing::{error, trace, Instrument};
+
+use super::{CreateConsumer, ScopedMessage};
+
+/// A convenience type for returning work results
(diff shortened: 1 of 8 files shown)


thomastaylor312:
Until #60 merges, just filter this PR to the single commit to see what I did

thomastaylor312:
Ignore build failures since I accidentally started with this against main

src/consumers/manager.rs
```diff
@@ -0,0 +1,168 @@
...
+    ///
+    /// The given work function should attempt to handle the event
+    pub async fn add_for_lattice<W>(&self, topic: &str, worker: W) -> Result<(), async_nats::Error>
+    where
+        W: Worker + Send + Sync + 'static,
+        C: Stream<Item = Result<ScopedMessage<W::Message>, async_nats::Error>>
+            + CreateConsumer<Output = C>
+            + Send
+            + Unpin
+            + 'static,
```
brooksmtownsend, line 107: tenor-108401073

brooksmtownsend (approved):
So this PR introduces the idea of "work" which I understand to be processing an event or command. As these are picked up on the NATS topic, we use a semaphore and create a work handler to represent the work needed to handle that event/command. With a slow stream of these this won't really change much, but with a large stream of events or commands the ConsumerManager will essentially deal with dealing out permits to let them execute.

Is this a decent summary? If this is way off the mark, then my approval may be moot

thomastaylor312:
Yep, you have it exactly right

Issue #45, opened 2023-03-04
Create storage struct for storing lattice state in NATS KV


thomastaylor312:
*   This solution should define how we store the data in a KV bucket (key structure, data encoding)
    *   Suggestion: Look at what the lattice observer does. Don't have to follow it exactly, but probably will be very similar
*   There should be a struct that auto creates a NATS KV bucket if it doesn't exists and handles all CRUD operations for state

PR #62, opened 2023-03-07
feat: track lattice state in kv

vados-cosmonic:
## Feature or Problem

Add code to support storing state (in a somewhat general fashion), and more specifically lattice state in NATS KV.

## Related Issues

#45 

## Release Information

`next`

## Consumer Impact

Consumers will be able to use a `wadm` flow that relies on NATS KV state.

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)

TBD

### Acceptance or Integration

TBD

### Manual Verification

TBD

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,51 @@
+name: Test Wadm (v0.4)
+
+on:
+  pull_request:
+    branches: [ wadm_0.4 ]
+
+jobs:
+  test:
+    name: Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-22.04 ]
+        nats_version: [ 2.9.15 ]
+
+    services:
+      nats:
+        image: ${{ matrix.nats_version }}-alpine
+        env:
+          JETSTREAM: ""
+
+    steps:
+      - uses: actions/checkout@v2
+
+      # Cache: rust
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: "${{ matrix.os }}-rust-cache"
+
+      # Cache: nats-server binary
+      - name: Cache nats binary
+        id: cache-nats
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-nats-binary
+        with:
+          path: /usr/bin/nats
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ matrix.nats_version }}
+
+      # Install: nats-serveer
+      - name: Install nats binary
+        if: ${{ steps.cache-nats.outputs.cache-hit != 'true' }}
+        run: |
+          curl -L https://github.com/nats-io/nats-server/releases/download/v${{ matrix.nats_version }}/nats-server-v${{ matrix.nats_version }}-linux-amd64.zip -o nats-server.zip
+          unzip nats-server.zip -d nats-server
+          sudo cp nats-server/nats-server-v${{ matrix.nats_version }}-linux-amd64/nats-server /usr/bin
+
+      # Run all integration tests
+      - name: Run integration tests
+        run: |
+          make test-int
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
 
     services:
       nats:
-        image: ${{ matrix.nats_version }}-alpine
+        image: nats:${{ matrix.nats_version }}-alpine
         env:
           JETSTREAM: ""
 
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ test-watch: ## Run tests (continuously)
 
 test-int:: ## Run integration tests
 ifeq (,$(CARGO_TEST_TARGET))
-	$(CARGO) test -- --nocapture
+	$(CARGO) test --test ="*" -- --nocapture
 else
 	$(CARGO) test --test $(CARGO_TEST_TARGET) -- --nocapture
 endif
--- a/Makefile
+++ b/Makefile
@@ -39,17 +39,17 @@ endif
 # Build #
 #########
 
-lint: check-cargo-clippy
+lint: check-cargo-clippy ## Run code lint
 	$(CARGO) fmt --all --check
 	$(CARGO) clippy --all-features --all-targets --workspace
 
-lint-watch: check-cargo-clippy
+lint-watch: check-cargo-clippy ## Run code lint (continuously)
 	$(CARGO) watch -- $(MAKE) lint
 
 build: ## Build wadm
 	$(CARGO) build
 
-build-watch: check-cargo-watch ## Build wadm continuously
+build-watch: check-cargo-watch ## Build wadm (continuously)
 	$(CARGO) watch -- $(MAKE) build
 
 ########
@@ -69,16 +69,20 @@ else
 	$(CARGO) test $(CARGO_TEST_TARGET) -- --nocapture
 endif
 
-test-watch:
+test-watch: ## Run tests (continuously)
 	$(CARGO) watch -- $(MAKE) test
 
-test-int:
+test-int:: ## Run integration tests
+ifeq (,$(CARGO_TEST_TARGET))
+	$(CARGO) test -- --nocapture
+else
 	$(CARGO) test --test $(CARGO_TEST_TARGET) -- --nocapture
+endif
 
-test-int-all:
-	$(MAKE) test-int CARGO_TEST_TARGET='*'
-
-test-int-watch:
+test-int-watch: ## Run integration tests (continuously)
 	$(CARGO) watch -- $(MAKE) test-int
+
+test-int-all:: ## Run all integration tests
+	$(MAKE) test-int CARGO_TEST_TARGET='*'
 
 .PHONY: check-cargo-watch check-cargo-clippy lint build build-watch test test-watch test-int test-int-all test-int-watch
--- a/Makefile
+++ b/Makefile
@@ -1,21 +1,84 @@
+SHELL := bash
+.ONESHELL:
+.SHELLFLAGS := -eu -o pipefail -c
+.DELETE_ON_ERROR:
+MAKEFLAGS += --warn-undefined-variables
+MAKEFLAGS += --no-builtin-rules
+MAKEFLAGS += --no-print-directory
+MAKEFLAGS += -S
+
 .DEFAULT: all
+
+CARGO ?= cargo
+CARGO_WATCH ?= cargo-watch
+CARGO_CLIPPY ?= cargo-clippy
+DOCKER ?= docker
 
 all: build test
 
-build:
-	cargo build --features cli
+help:  ## Display this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_\-.*]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-ifeq ($(shell nc -czt -w1 127.0.0.1 4222 || echo fail),fail)
-test::
-	docker run --rm -d --name wadm-test -p 127.0.0.1:4222:4222 nats:2.9 -js
-	$(MAKE) test
-	docker stop wadm-test
-else
-test:: _test
+###########
+# Tooling #
+###########
+
+# Ensure that cargo watch is installed
+check-cargo-watch:
+ifeq ("",$(shell command -v $(CARGO_WATCH)))
+	$(error "ERROR: cargo-watch is not installed (see: https://crates.io/crates/cargo-watch)")
 endif
 
-# Internal test target for reuse
-_test:
-	cargo test -- --nocapture
+# Ensure that clippy is installed
+check-cargo-clippy:
+ifeq ("",$(shell command -v $(CARGO_CLIPPY)))
+	$(error "ERROR: clippy is not installed (see: https://doc.rust-lang.org/clippy/installation.html)")
+endif
 
-.PHONY: test build all
+#########
+# Build #
+#########
+
+lint: check-cargo-clippy
+	$(CARGO) fmt --all --check
+	$(CARGO) clippy --all-features --all-targets --workspace
+
+lint-watch: check-cargo-clippy
+	$(CARGO) watch -- $(MAKE) lint
+
+build: ## Build wadm
+	$(CARGO) build
+
+build-watch: check-cargo-watch ## Build wadm continuously
+	$(CARGO) watch -- $(MAKE) build
+
+########
+# Test #
+########
+
+# An optional specific test for cargo to target
+
+CARGO_TEST_TARGET ?=
+
+test:: ## Run tests
+ifeq ($(shell nc -czt -w1 127.0.0.1 4222 || echo fail),fail)
+	$(DOCKER) run --rm -d --name wadm-test -p 127.0.0.1:4222:4222 nats:2.9 -js
+	$(CARGO) test $(CARGO_TEST_TARGET) -- --nocapture
+	$(DOCKER) stop wadm-test
+else
+	$(CARGO) test $(CARGO_TEST_TARGET) -- --nocapture
+endif
+
+test-watch:
+	$(CARGO) watch -- $(MAKE) test
+
+test-int:
+	$(CARGO) test --test $(CARGO_TEST_TARGET) -- --nocapture
+
+test-int-all:
+	$(MAKE) test-int CARGO_TEST_TARGET='*'
+
+test-int-watch:
+	$(CARGO) watch -- $(MAKE) test-int
+
+.PHONY: check-cargo-watch check-cargo-clippy lint build build-watch test test-watch test-int test-int-all test-int-watch
--- a/bin/main.rs
+++ b/bin/main.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use async_nats::jetstream::{self, stream::Config, Context};
 use clap::Parser;
+use std::path::PathBuf;
 use tokio::sync::Semaphore;
 use tracing::{debug, error, trace};
 
@@ -13,7 +14,11 @@ use wadm::{
     commands::*, events::*, DEFAULT_COMMANDS_TOPIC, DEFAULT_EVENTS_TOPIC, DEFAULT_EXPIRY_TIME,
 };
 
+use wadm::storage::{NatsAuthConfig, NatsKvStorageConfig, NatsKvStorageEngine};
+
 mod logging;
+
+const DEFAULT_STORAGE_NKV_URL: &str = "localhost:4222";
 
 #[derive(Parser, Debug)]
 #[command(name = clap::crate_name!(), version = clap::crate_version!(), about = "wasmCloud Application Deployment Manager", long_about = None)]
@@ -75,6 +80,47 @@ struct Args {
         env = "WADM_MAX_JOBS"
     )]
     max_jobs: usize,
+
+    /// URL of the Nats Jetstream KV that will be used for storage
+    #[arg(long = "storage-nkv-url", env = "STORAGE_NATSKV_URL")]
+    storage_nkv_url: Option<String>,
+
+    /// Optional prefix for the server tate that will be stored.
+    /// this prefix will be prepended to the bucket name of the lattice
+    /// (ex. <prefix>_lattice_<lattice id>).
+    #[arg(
+        long = "storage-nkv-lattice-bucket-prefix",
+        env = "WADM_STORAGE_NATSKV_LATTICE_BUCKET_PREFIX"
+    )]
+    storage_nkv_lattice_bucket_prefix: Option<String>,
+
+    /// (Optional) NATS NKey authentication
+    #[arg(long = "storage-nkv-nkey", env = "WADM_STORAGE_NATSKV_NKEY")]
+    storage_nkv_nats_auth_nkey: Option<String>,
+
+    /// (Optional) NATS credential file to use when authenticating
+    #[arg(
+        long = "storage-nkv-nats-auth-creds-file",
+        env = "WADM_STORAGE_NATSKV_NATS_AUTH_CREDS_FILE",
+        conflicts_with = "storage_nkv_nats_auth_jwt_seed"
+    )]
+    storage_nkv_nats_auth_creds_file: Option<String>,
+
+    /// (Optional) NATS JWT seed to use when authenticating
+    #[arg(
+        long = "storage-nkv-nats-auth-jwt-seed",
+        env = "WADM_STORAGE_NATSKV_NATS_AUTH_JWT_SEED",
+        conflicts_with = "storage_nkv_nats_auth_creds_file"
+    )]
+    storage_nkv_nats_auth_jwt_seed: Option<String>,
+
+    /// (Optional) NATS JWT file to use when authenticating
+    #[arg(
+        long = "storage-nkv-nats-auth-jwt-path",
+        env = "WADM_STORAGE_NATSKV_NATS_AUTH_JWT_PATH",
+        conflicts_with = "storage_nkv_nats_auth_creds_file"
+    )]
+    storage_nkv_nats_auth_jwt_path: Option<String>,
 }
 
 #[tokio::main]
@@ -86,6 +132,24 @@ async fn main() -> anyhow::Result<()> {
         args.tracing_enabled,
         args.tracing_endpoint,
     );
+
+    // Build storage adapter for lattice state (on by default)
+    let natskv_storage_config = NatsKvStorageConfig {
+        nats_url: args
+            .storage_nkv_url
+            .unwrap_or(DEFAULT_STORAGE_NKV_URL.into()),
+        lattice_bucket_prefix: args.storage_nkv_lattice_bucket_prefix,
+        auth: Some(NatsAuthConfig {
+            creds_file: args.storage_nkv_nats_auth_creds_file,
+            jwt_seed: args.storage_nkv_nats_auth_jwt_seed,
+            jwt_path: args.storage_nkv_nats_auth_jwt_path.map(PathBuf::from),
+        }),
+    };
+
+    // TODO: use the storage engine
+    let _storage = NatsKvStorageEngine::new(natskv_storage_config)
+        .await
+        .map_err(|e| anyhow::anyhow!("{e:?}"))?;
 
     // TODO: All the NATS connection options and jetstream stuff
     let client = async_nats::connect("127.0.0.1:4222").await?;
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -1,0 +1,106 @@
+use async_trait::async_trait;
+use semver::Version;
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+use thiserror::Error;
+
+mod nats_kv;
+mod state;
+
+pub use nats_kv::{NatsAuthConfig, NatsKvStorageConfig, NatsKvStorageEngine};
+pub use state::{Actor, Claim, Host, LatticeParameters, LatticeState, Provider};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct WithStateMetadata<T> {
+    /// Revision (if supported), often used for compare-and-set-semantics
+    pub revision: Option<u64>,
+
+    /// The state that was returned
+    pub state: T,
+}
+
+/////////////
+// Engines //
+/////////////
+
+/// Pluggable engines can be asked for their metadata
+/// to enable generic (possibly dyn) usage
+#[async_trait]
+pub trait EngineMetadata {
+    /// Get the name of the lattice storage
+    async fn name() -> String;
+
+    /// Get the version of the lattice storage
+    async fn version() -> Version;
+}
+
+/////////////////////
+// Storage Engines //
+/////////////////////
+
+/// Errors that are related to storage engines
+#[derive(Debug, Error)]
+pub enum StorageEngineError {
+    /// When the underlying engine throws an error
+    #[error("Storage engine encountered an error: {0:?}")]
+    Engine(Box<dyn Debug + Send + Sync>),
+
+    /// An error when the cause is unknown. This happens very rarely and should be considered fatal
+    #[error("Unknown error has occured")]
+    Unknown,
+}
+
+/////////////
+// Storage //
+/////////////
+
+/// Errors that are related to storage
+#[derive(Debug, Error)]
+pub enum StorageError {
+    /// Errors that arise at the engine layer
+    #[error("Storage error: {0}")]
+    Engine(String),
+
+    /// Errors that arise at the network layer
+    #[error("Network error: {0}")]
+    NetworkError(String),
+
+    /// An error when the cause is unknown. This happens very rarely and should be considered fatal
+    #[error("Unknown error has occured: {0}")]
+    Unknown(String),
+}
+
+/// Options and metadata that can be used to improve store operations
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct StoreOptions {
+    /// Whether to force a store with compare-and-set semantics
+    pub require_cas: bool,
+
+    /// version identifier that enables stores to perform compare-and-set semantics
+    pub revision: Option<u64>,
+}
+
+/// A trait that indicates the ability of a struct to store state
+/// Objects that implement Store can save some given State, identifiable by StateId, and updatable with the given StateBuilder
+///
+#[async_trait]
+pub trait Store {
+    type State;
+    type StateId;
+
+    /// Get a particular piece of state
+    async fn get(&self, id: Self::StateId) -> Result<WithStateMetadata<Self::State>, StorageError>;
+
+    /// Store a piece of state
+    async fn store(
+        &mut self,
+        state: Self::State,
+        opts: StoreOptions,
+    ) -> Result<WithStateMetadata<Self::StateId>, StorageError>;
+
+    /// Delete an existing piece of state
+    async fn delete(&self, id: String) -> Result<(), StorageError>;
+}
+
+/// LatticeStorages are state stores that support storing lattice information
+pub trait LatticeStorage: Store<State = LatticeState, StateId = String> {}
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -20,23 +20,8 @@ pub struct WithStateMetadata<T> {
 }
 
 /////////////
-// Engines //
+// Storage //
 /////////////
-
-/// Pluggable engines can be asked for their metadata
-/// to enable generic (possibly dyn) usage
-#[async_trait]
-pub trait EngineMetadata {
-    /// Get the name of the lattice storage
-    async fn name() -> String;
-
-    /// Get the version of the lattice storage
-    async fn version() -> Version;
-}
-
-/////////////////////
-// Storage Engines //
-/////////////////////
 
 /// Errors that are related to storage engines
 #[derive(Debug, Error)]
@@ -49,10 +34,6 @@ pub enum StorageEngineError {
     #[error("Unknown error has occured")]
     Unknown,
 }
-
-/////////////
-// Storage //
-/////////////
 
 /// Errors that are related to storage
 #[derive(Debug, Error)]
@@ -81,12 +62,17 @@ pub struct StoreOptions {
 }
 
 /// A trait that indicates the ability of a struct to store state
-/// Objects that implement Store can save some given State, identifiable by StateId, and updatable with the given StateBuilder
-///
+/// Objects that implement Store can retrieve, save and delete some given State, identifiable by StateId
 #[async_trait]
 pub trait Store {
     type State;
     type StateId;
+
+    /// Get the name of the lattice storage
+    async fn name() -> String;
+
+    /// Get the version of the lattice storage
+    async fn version() -> Version;
 
     /// Get a particular piece of state
     async fn get(&self, id: Self::StateId) -> Result<WithStateMetadata<Self::State>, StorageError>;
--- a/src/storage/nats_kv.rs
+++ b/src/storage/nats_kv.rs
@@ -18,7 +18,7 @@ use serde::ser::StdError;
 use serde::{Deserialize, Serialize};
 use serde_json::{from_slice, to_vec, Error as SerdeJsonError};
 
-use super::{EngineMetadata, LatticeStorage, StorageEngineError, StorageError, StoreOptions};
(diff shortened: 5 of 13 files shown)


src/storage/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+};
+
+/////////////
+// Engines //
+/////////////
+
+pub type EngineName = String;
+
+/// An engine for performing some task that is pluggable
+pub trait Engine {
```
thomastaylor312, line 22: Nit: maybe EngineMetadata

src/storage/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    /// An error when the cause is unknown. This happens very rarely and should be considered fatal
+    #[error("Unknown error has occured: {0}")]
+    Unknown(String),
+}
+
+/// A trait that indicates the ability of a struct to store state
+/// Objects that implement Stores can save some given State, identifiable by StateId, and updatable with the given StateBuilder
+#[async_trait]
+pub trait Stores<State, StateId, StateBuilder> {
```
thomastaylor312, line 86: In Rust it is better to do these with associated types (e.g. `type State` and then referencing with `Self::State` in the functions

src/storage/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+    // Get a particular piece of state
+    async fn get(id: StateId) -> Result<State, StorageError>;
+
+    // Store a piece of state
+    async fn store(state: State) -> Result<State, StorageError>;
+
+    // Update an existing piece of state
+    async fn update(id: StateId, update: StateBuilder) -> Result<(), StorageError>;
+}
```
thomastaylor312, line 95: Missing delete?

src/storage/nats_kv.rs
```diff
@@ -0,0 +1,106 @@
...
+        NatsKvStorageEngine {
+            nats_url,
+            client: None,
+        }
+    }
+}
+
+impl Engine for NatsKvStorageEngine {
+    fn name() -> EngineName {
+        String::from(STORAGE_ENGINE_NAME)
```
thomastaylor312, line 61: Nit: in these functions it is better to do `to_owned()` on them. But it is making me think that maybe these functions should return &str

thomastaylor312 (commented):
Great start here!

src/storage/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+/////////////
+// Engines //
+/////////////
+
+pub type EngineName = String;
+
+/// An engine for performing some task that is pluggable
+pub trait Engine {
+    /// Get the name of the lattice storage
+    fn name() -> EngineName;
```
thomastaylor312, line 24: One more idea I just had: maybe make these pub consts rather than functions. See what I did with events

src/storage/nats_kv.rs
vados-cosmonic: I'm trying to prep for a world where the Engine on the other side (maybe I should have made this `async` to show that) could be completely remote (and thus we couldn't have the version as a `&str` per-say). i.e. `RemoteStorageEngine`.

src/storage/mod.rs
vados-cosmonic: my note above about being able to use a `RemoteStorageEngine` -- I want to be able to ask a storage engine it's name (and it possibly do some work that I don't know about to find out). I didn't go as far as making this `async name()`, but I kind of want to

src/storage/mod.rs
vados-cosmonic: Yup! it'll be in there soon, don't worry, just wanted to get something up review ASAP

src/storage/mod.rs
vados-cosmonic: Sure, will change this! These types are certainly unlikely to vary and are related, so associated types make sense here

src/storage/mod.rs
vados-cosmonic: [added!](https://github.com/wasmCloud/wadm/pull/62/commits/617e1d78d1849fc8ea129048b2fac6ac086cf38c)

src/storage/mod.rs
vados-cosmonic: [changed!](https://github.com/wasmCloud/wadm/pull/62/commits/52bb8baf091d850fb380040eff25ff460e056070)

src/storage/mod.rs
vados-cosmonic: [changed!](https://github.com/wasmCloud/wadm/pull/62/commits/52bb8baf091d850fb380040eff25ff460e056070)

src/storage/nats_kv.rs
thomastaylor312: That is an interesting use case. Probably keep it for now

src/storage/nats_kv.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// NATS Jetstream KV Store
+    kv_store: Option<JetstreamKvStore>,
+}
+
+impl NatsKvStorageEngine {
+    fn new(nats_url: String) -> NatsKvStorageEngine {
+        NatsKvStorageEngine {
+            nats_url,
+            nats_prefix: env::var(ENV_VAR_STORAGE_NATS_KV_PREFIX).unwrap_or(String::from(DEFAULT_NATS_KV_PREFIX)),
```
thomastaylor312, line 68: These should be config items passed in to the struct rather than env vars here. I just merged #60 that has the flags and stuff set up to get the data!

src/storage/nats_kv.rs
```diff
@@ -0,0 +1,187 @@
...
+            client: None,
```
thomastaylor312, line 69: I would set up the client in `new` rather than using the trait to connect. It’ll keep it simpler for now. Also avoids the need to unwrap the option every time. If we need to change it later we can

src/storage/state.rs
```diff
@@ -0,0 +1,96 @@
...
+////////////
+// Claims //
+////////////
+
+/// A Claim that can be applied to a given lattice
+#[derive(Debug,Serialize, Deserialize, Clone)]
+pub struct Claim {
+    pub name: String,
+
+    pub version: String, /// TODO(QUESTION): Do we enforce semver?
```
thomastaylor312, line 17: To answer your question: No, this is just a usize

vados-cosmonic: I thought the `revision` was  a `usize` and the `version` is a string? That's how it is in [`events/data.rs`](https://github.com/wasmCloud/wadm/blob/wadm_0.4/src/events/data.rs#L49)

src/storage/nats_kv.rs
vados-cosmonic: Agreed, will make this change@

- [ ] Set up client connection in `new`, remove `Option`

src/storage/nats_kv.rs
vados-cosmonic: Yup! agreed

- [ ] use config object
- [ ] integrate changes from #60, add flags as necessary

src/storage/mod.rs
vados-cosmonic: Resolving this due to the other discussion, slightly overkill, but enables some cool future use cases IMO!

src/storage/state.rs
thomastaylor312: Oh sorry you're correct. I was looking at this on phone and misread. To answer the todo question, we do expect this field to be semver

src/storage/mod.rs
```diff
@@ -78,10 +69,37 @@ pub enum StorageError {
...
+/// Trait for objects that can read and write some piece of state with a given identifier
+#[async_trait]
+pub trait StoreReadWrite<State,Id>: StoreRead<State,Id> + StoreWrite<State,Id> {}
+
 /// A trait that indicates the ability of a struct to store state
 /// Objects that implement Stores can save some given State, identifiable by StateId, and updatable with the given StateBuilder
+///
 #[async_trait]
-pub trait Stores {
+pub trait Stores
```
thomastaylor312, line 98: Nit: I think most traits are singular

src/storage/mod.rs
```diff
@@ -78,10 +69,37 @@ pub enum StorageError {
...
+pub trait StoreRead<State,Id> {
+    /// Read a given piece of state from the store
+    async fn read(&self, id: Id) -> Result<State, StorageError>;
+}
+
+/// A trait that represents whether some struct can be written to a given store
+#[async_trait]
+pub trait StoreWrite<State,Id> {
+    /// Write to a given store
+    async fn write(&mut self, obj: State, id: Option<Id>) -> Result<Id, StorageError>;
```
thomastaylor312, line 87: Is there a reason why we wouldn't ever need an ID? Feels like a cleaner API for this not to be an option

src/storage/nats_kv.rs
```diff
@@ -3,8 +3,8 @@
 //! This storage engine enables storing any WADM related information using NATS Kv as a backend
 
 use std::env;
-use std::sync::Arc;
-use std::sync::RwLock;
+use std::sync::{Arc, RwLock};
```
thomastaylor312, line 6: I didn't notice this during the first go around, but this should probably use `tokio::sync::RwLock`. It is nicer to use (no result type) and built to work with our async code

src/storage/mod.rs
vados-cosmonic: I'm good with that!

- [ ] Change `Stores` to `Store`

src/storage/mod.rs
vados-cosmonic: This was in the case that you *don't know* the ID ahead of time -- like if you're storing something and you're going to get your ID from the storage layer.

Could be premature, since just about every storage layer will probably be happy with a uuidv1/4/7/8

src/storage/nats_kv.rs
vados-cosmonic: So I actually took out that stuff in favor of just cloning the `Client` since it's already there, when someone needs a KV bucket.

It doesn't look like creating the KV bucket references is actually a heavy operation

src/storage/state.rs
vados-cosmonic: Thanks, will change this to a `semver::Version` then!

.github/workflows/test.yml
```diff
@@ -0,0 +1,46 @@
...
+  pull_request:
+    branches: [ wadm_0.4 ]
+
+jobs:
+  test:
+    name: Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-20.04 ]
```
rvolosatovs, line 13:
```suggestion
        os: ubuntu-22.04
```

1. There's only one OS specified, so no need for the list
2. `22.04` is the latest, unless there is a reason to use `20.04`, which would need to be documented here

bin/main.rs
```diff
@@ -77,6 +84,12 @@ struct Args {
     /// The NATS JetStream domain to connect to
     #[arg(short = 'd', env = "WADM_JETSTREAM_DOMAIN")]
     domain: Option<String>,
+
+    #[arg(long = "storage-nkv-url", env = "STORAGE_NATSKV_URL")]
+    storage_nkv_url: Option<String>,
+
+    #[arg(long = "storage-nkv-prefix", env = "STORAGE_NATSKV_PREFIX")]
```
autodidaddict, line 91: we should probably add some help text to this. Not everyone will know what this prefix is or how it impacts the deployment.

Makefile
```diff
@@ -1,21 +1,77 @@
...
-_test:
-	cargo test -- --nocapture
+test-watch:
+	$(CARGO) watch -- $(MAKE) test
+
+test-int:
+	$(CARGO) test --test '*' -- --nocapture
+
+test-int-watch:
+	$(CARGO) watch -- $(MAKE) test-int
```
rvolosatovs, line 75: Maybe it's just me, but all of this looks like scripts that belong in developers local environment and not a shared project `Makefile`

Makefile
```diff
@@ -1,21 +1,77 @@
...
+	$(CARGO) build
+
+build-watch: check-cargo-watch ## Build wadm continuously
+	$(CARGO) watch -- $(MAKE) build
+
+########
+# Test #
+########
+
+# An optional specific test for carog to target
```
rvolosatovs, line 56:
```suggestion
# An optional specific test for cargo to target

```

bin/main.rs
```diff
@@ -305,13 +327,13 @@ fn configure_tracing(
         .with(log_layer);
     if !tracing_enabled && tracing_endpoint.is_none() {
         if let Err(e) = tracing::subscriber::set_global_default(subscriber) {
-            eprintln!("Logger/tracer was already initted, continuing: {}", e);
+            eprintln!("Logger/tracer was already initted, continuing: {e}");
```
autodidaddict, line 330: grammar nit (super tiny): should probably use `initialized` here instead of `initted`. Less likely to cause confusion for non-native speakers.

Makefile
```diff
@@ -1,21 +1,77 @@
...
+CARGO_TEST_TARGET ?=
+
+test:: ## Run tests
 ifeq ($(shell nc -czt -w1 127.0.0.1 4222 || echo fail),fail)
-test::
-	docker run --rm -d --name wadm-test -p 127.0.0.1:4222:4222 nats:2.9 -js
-	$(MAKE) test
-	docker stop wadm-test
+	$(DOCKER) run --rm -d --name wadm-test -p 127.0.0.1:4222:4222 nats:2.9 -js
+	$(CARGO) test $(CARGO_TEST_TARGET) -- --nocapture
```
rvolosatovs, line 62: Why hard-code `--nocapture`? Why not stick with defaults and let the developer choose the arguments like one would do with a
```
cargo test ${@}
```
in a bash script

Makefile
```diff
@@ -1,21 +1,77 @@
...
 
+lint: check-cargo-clippy
+	$(CARGO) fmt --all --check
+	$(CARGO) clippy --all-features --all-targets --workspace
+
+build: ## Build wadm
+	$(CARGO) build
+
+build-watch: check-cargo-watch ## Build wadm continuously
+	$(CARGO) watch -- $(MAKE) build
```
rvolosatovs, line 50: Is a Makefile directive really required for this?

src/storage/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// Whether to force a store with compare-and-set semantics
+    pub require_cas: bool,
+
+    /// version identifier that enables stores to perform compare-and-set semantics
+    pub revision: Option<u64>,
+}
+
+
+
+/// A trait that represents whether some struct can be read from a given store
```
autodidaddict, line 101: This isn't something that needs to be changed, but is more of an architectural design preference for evolutionary (rather than predictive) architectures. Here it feels like we've made this code a bit more difficult to read in order to support multiple storage engines in the future. In the interest of the  "keep it simple" principle the code would probably be simpler if we just had a single storage engine for now and extrapolated to generic types later as needed.

Again, not something that I think prevents the PR from being accepted, just want to be explicit that we're incurring some readability debt here.

tests/storage_nats_kv.rs
```diff
@@ -0,0 +1,156 @@
+use anyhow::Result;
+use std::collections::HashMap;
+
+mod common;
+use common::make_nats_instance;
+use semver::Version;
+
+use wadm::storage::{
```
rvolosatovs, line 8: spacing here is inconsistent

rvolosatovs (commented):
I took a quick look, mostly at tooling, will dive deeper later

src/storage/nats_kv.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    /// Prefix to prepend to NATS bucket name
+    pub nats_prefix: Option<String>,
+}
+
+impl NatsKvStorageEngine {
+    #[allow(dead_code)]
+    pub async fn new(config: NatsKvStorageConfig) -> Result<NatsKvStorageEngine, StorageEngineError> {
+        // Connect to NATS
+        let client = async_nats::connect(&config.nats_url).await
```
autodidaddict, line 80: This should probably be upgraded to support actual authentication. This code will only connect to anonymously authenticated servers, whereas most other wasmCloud components that talk to NATS support authenticating with the JWT+seed combo

src/storage/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+///
+/// Note that these msut be this way because we can't write non-overlapping blanked impls yet:
```
rvolosatovs, line 103:
```suggestion
/// Note that these must be this way because we can't write non-overlapping blanked impls yet:
```

src/storage/nats_kv.rs
```diff
@@ -0,0 +1,248 @@
...
+        Ok(NatsKvStorageEngine {
+            config,
+            client,
+        })
+    }
+
+    /// Build the top level bucket for the lattice
+    fn build_lattice_bucket_name(&self, lattice_id: String) -> String {
+        let mut bucket = format!("lattice_{lattice_id}");
+        if let Some(prefix) = &self.config.nats_prefix {
```
autodidaddict, line 92: Mentioned this in another file but is there something we can call `nats_prefix` to help disambiguate it from other prefixes we typically use in wamCloud code (e.g. `lattice prefix`)

src/storage/nats_kv.rs
```diff
@@ -0,0 +1,248 @@
...
+        bucket
+    }
+
+    /// Get or create the KV store for NATS jetstream
+    async fn create_kv_store(&self, bucket: String) -> Result<JetstreamKvStore, StorageError> {
+        // Build jetstream client
+        let jetstream = async_nats::jetstream::new(self.client.clone());
+
+        // Create Jetstream KV store
+        let kv = jetstream.create_key_value(JetstreamKvConfig {
```
autodidaddict, line 104: We should upgrade this to check for the existence of the key value bucket. As far as I've been able to test, the only time `create_key_value` passes when an existing bucket is already there is if the configuration is exactly the same.

If we don't do a get first to check for a pre-existing one, then this code will fail if someone has created a key value bucket already with a custom replication and clustering or persistence configuration.

autodidaddict: From some other code that I've been working on, this attempts to get and, if not found, creates a new one.
```rust

async fn get_or_create_bucket(js: &Context) -> Result<Store> {
    if let Ok(store) = js
        .get_key_value(STATE_BUCKET_NAME)
        .await
        .map_err(|e| RpcError::Nats(e.to_string()))
    {
        Ok(store)
    } else {
        Ok(js
            .create_key_value(KvConfig {
                bucket: STATE_BUCKET_NAME.to_string(),
                description: "my state bucket".to_string(),
                history: 10,
                ..Default::default()
            })
            .await
            .map_err(|e| RpcError::Nats(e.to_string()))?)
    }
}
```

src/storage/nats_kv.rs
```diff
@@ -0,0 +1,248 @@
...
+/// For every lattice this implementation stores the all lattice state under a single key
+/// determined by LATTICE_BUCKET_STATE_KEY.
+/// For example, a lattice bucket "lattice_1", you can find serialized lattice state at "lattice_1.state".
+#[async_trait]
+impl StoreRead<LatticeState,String> for NatsKvStorageEngine {
+    async fn read(&self, id: String) -> Result<WithStateMetadata<LatticeState>, StorageError> {
+        let bucket = self.build_lattice_bucket_name(id);
+        let kv = self.create_kv_store(bucket.clone()).await?;
+
+        let lattice;
```
autodidaddict, line 195: does Clippy actually let you get away with this? I remember doing this once and Clippy got super mad at me

tests/common.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+/// A process-based NATS instance
+struct ProcessNats {
+    child: Child,
+    port: u16,
+}
+
+impl ProcessNats {
```
autodidaddict, line 90: In other tests that require NATS, we've just ensured or assumed that NATS was running. For example, in the wasmCloud OTP host, the CI scripts start nats as a service and the scripts to run individual tests.

It feels like a lot of extra complexity (and potentially brittle execution inside the github CI) to try and manually launch nats within the test code when it should be considerably easier to just let the environment take care of starting NATS.

Thoughts on this @thomastaylor312 ?

tests/common.rs
```diff
@@ -0,0 +1,197 @@
...
+    id: String,
+
+    // Container name
+    container_name: String
+}
+
+const CONTAINER_SUFFIX_ALPHABET: &str = "01234567890";
+
+impl DockerizedNats {
+    async fn new(port: u16) -> Result<DockerizedNats> {
```
autodidaddict, line 147: see my previous comment. It feels like we're incurring a pile of overhead in launching docker and NATS etc when environment scripts like Makefiles, bash scripts, and github workflows should be doing this

autodidaddict (changes requested):
Just a few questions and an overall (non-blocking) comment that it feels like we may be doing too much work here both in terms of the abstraction and on our tests where we're launching NATS by hand.

Main request change is adding support for JWT authentication to the NATS client config

tests/common.rs
vados-cosmonic: The thinking here was that using the methods up til now require running serially -- since the NATS is shared between tests they have to run serially. In general the only way past this is running setup per-test or per-suite.

The reason I needed two ways to do this was because Github CI doesn't have docker dind (otherwise it could be consistently docker across local and CI environments and a lot more "contained") -- also the alternative was nice for people who don't like using `docker`.

.github/workflows/test.yml
vados-cosmonic: Hey thanks for the feedback -- I don't think that there *should* be only one OS for now (the other `wadm` workflows have more), and the other wadm tests use `20.04` rather than `22.04` which is why I did the same thing here. 

Super open to changing this though -- @thomastaylor312  @autodidaddict can I update this bit?

.github/workflows/test.yml
```diff
@@ -0,0 +1,46 @@
...
+        nats_version: [2.9.15]
```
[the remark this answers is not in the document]
vados-cosmonic: Yeah I picked the latest version of NATS to match the default, but there's definitely only one of those, will fix

- [ ] change `nats_version` to only one version

Makefile
```diff
@@ -1,21 +1,77 @@
...
+CARGO_WATCH ?= cargo-watch
+CARGO_CLIPPY ?= cargo-clippy
+DOCKER ?= docker
+
 all: build test
 
-build:
-	cargo build --features cli
+help:  ## Display this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_\-.*]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```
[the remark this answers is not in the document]
vados-cosmonic: This is a `Makefile` formula for printing help text and is pretty common across all the wasmCloud projects, and is generally pretty helpful because it's a facility Makefiles lack. 

It'd be great if we could move to `Justfile`s in the future (it's built-in there), but I think not having a `help` target would be a bad for ergonomics.

Makefile
vados-cosmonic: As far as `test-int` goes, I'd want these to be the tests that are run from CI, and be the source of truth for repeated commands.

The `-watch` files fit more in the creature comfort category as you're suggesting, and there I'd say that making it easy for people to have a tight dev loop is desirable.

One thing I think is missing from this is actually `test-int-all` versus `test-int` with a `CARGO_TEST_TARGET`, so someone could continuously test a *single* integration test.

Makefile
vados-cosmonic: While `$@` would work in a bash script, it doesn't work quite that way in `make` -- that's usually done with some sort of `ARGS` variable.

As for the `--nocapture` setting, I'm fine with removing it but this is basically what tests looked like across all of the wasmCloud codebase so far (probably to prevent surprise/skip a step when someone is `println!`-debugging a test). I'm fine with changing it, but was going for consistency/least surprise here

Makefile
vados-cosmonic: It's not required, but the important bit is to have a place to helpfully tell the user what is wrong and *how to fix it*. `check-cargo-watch` notes what/where `cargo-watch` is so that contributors can fix a little easier.

tests/storage_nats_kv.rs
vados-cosmonic: Ah will fix, meant to separate local module stuff and external deps!

- [ ] Fix grouping of external versus internal modules

bin/main.rs
vados-cosmonic: I think this was a clippy auto-fix, but will update the text here!

- [ ] "initted" -> "initialized"

bin/main.rs
vados-cosmonic: oversight on my part, will add some help text

- [ ] Add help text for `--storage-*` options

src/storage/mod.rs
vados-cosmonic: I think this *can* be somewhat mitigated with a file like `docs/storage/writing-a-storage-engine.md` or something of the sort (which has a minimal example), but yeah, I definitely agree with the loss of readability here...

src/storage/nats_kv.rs
vados-cosmonic: Yup, will update `NatsKvStorageConfig`

- [ ] Update `NatsKvStorageConfig` to support authenticated client connections (i.e. JWT + seed)

src/storage/nats_kv.rs
vados-cosmonic: yeah that's much better, will change.

- [ ] `nats_prefix` -> `lattice_bucket_prefix`
- [ ] Add specific explanation of bucket meaning to command line switch

src/storage/nats_kv.rs
vados-cosmonic: I had this earlier on... Can't remember why I removed it. Definitely should be a `get_or_create` rather than a `create`.  Will update this.

- [ ] use get_or_create pattern for bucket

src/storage/nats_kv.rs
vados-cosmonic: Yup! :) I added the `make lint` just so I could clippy. `fmt` has a bunch of complaints but clippy is happy

.github/workflows/test.yml
vados-cosmonic: OK going to just change this!

bin/main.rs
```diff
@@ -75,6 +102,48 @@ struct Args {
...
+    )]
+    storage_nkv_nats_auth_jwt_seed: Option<String>,
+
+    /// (Optional) NATS JWT file to use when authenticating
+    #[arg(
+        long = "storage-nkv-nats-auth-jwt-path",
+        env = "STORAGE_NATSKV_NATS_AUTH_JWT_PATH"
+    )]
+    storage_nkv_nats_auth_jwt_path: Option<String>,
+>>>>>>> 0907ff2 (feat: track lattice state in KV)
```
autodidaddict, line 146: Looks like there's a bit of a merge conflict still sitting in this file?

tests/common.rs
autodidaddict: When I need to run stuff serially, I just make sure that I run cargo test with a single test thread.

autodidaddict (changes requested):
Once the tests and build pass and what looks like a fragment of a merge conflict is cleaned up, then I'll defer to @thomastaylor312 on whether or not we should be launching NATS from directly within the Rust tests.

It still feels like this isn't something we should do, especially given that none of our other repositories do this for tests that require NATS.

bin/main.rs
vados-cosmonic: Ah sorry, just fixed this!

vados-cosmonic:
Hey @autodidaddict would you mind kicking off the build?

autodidaddict:
Trying to figure out how 😁

vados-cosmonic:
Oh you know what, I need to temporarily enable it to run here -- the tests normally wouldn't run on this branch's merge

vados-cosmonic:
Found the issue! That matrix syntax (when there's only one) doesn't work! it needs to be an array, or a [list of arrays](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) as I originally thought.

tests/common.rs
thomastaylor312: Yeah I already put the stuff that starts NATS for you if it isn't already running in the `Makefile`. Would prefer to keep that part there

src/storage/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+/////////////////////
+// Storage Engines //
+/////////////////////
+
+/// Errors that are related to storage engines
+#[derive(Debug, Error)]
+pub enum StorageEngineError {
+    /// When the underlying engine throws an error
+    #[error("Storage engine encountered an error: {0}")]
+    Engine(String),
```
thomastaylor312, line 41: After looking at the NATS stuff, this could be nicer if this was `T` where `T: Debug, Error` so we can have concrete error types.

src/storage/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    #[error("Storage engine [{0}] not connected")]
+    NotConnected(String),
```
thomastaylor312, line 44: Because we don't have a connect method any more, do we need this error? If so, should we just not have a top level error and instead allow a store to specify its own error type?

src/storage/nats_kv.rs
```diff
@@ -0,0 +1,304 @@
...
+        return NatsConnectOptions::with_credentials_file(path.clone().into())
+            .await
+            .map_err(|e| StorageEngineError::Engine(format!("Failed to connect NATS: {e}")));
+    }
+
+    Ok(NatsConnectOptions::default())
+}
+
+impl NatsKvStorageEngine {
+    #[allow(dead_code)]
```
thomastaylor312, line 108: I assume this was for testing? Should probably be removed now that this is `pub`

src/storage/nats_kv.rs
```diff
@@ -0,0 +1,304 @@
...
+impl From<std::io::Error> for StorageEngineError {
+    fn from(err: std::io::Error) -> StorageEngineError {
+        StorageEngineError::Engine(format!("unexpected I/O error: {err}"))
+    }
+}
+
+async fn build_nats_options(
+    cfg: &NatsAuthConfig,
+) -> Result<NatsConnectOptions, StorageEngineError> {
+    // Use JWT + seed if present
```
thomastaylor312, line 83: Super nit: Might be easier to rewrite this function to have a single match block with each arm returning the connect options

bin/main.rs
```diff
@@ -75,6 +102,47 @@ struct Args {
...
+    /// URL of the Nats Jetstream KV that will be used for storage
+    #[arg(long = "storage-nkv-url", env = "STORAGE_NATSKV_URL")]
+    storage_nkv_url: Option<String>,
+
+    /// Optional prefix for the server tate that will be stored.
+    /// this prefix will be prepended to the bucket name of the lattice
+    /// (ex. <prefix>_lattice_<lattice id>).
+    #[arg(
+        long = "storage-nkv-lattice-bucket-prefix",
+        env = "STORAGE_NATSKV_LATTICE_BUCKET_PREFIX"
```
thomastaylor312, line 115: Generally environment vars should be prefaced with the application name (i.e. `WADM_`)

bin/main.rs
```diff
@@ -75,6 +102,47 @@ struct Args {
...
+    )]
+    storage_nkv_lattice_bucket_prefix: Option<String>,
+
+    /// (Optional) NATS NKey authentication
+    #[arg(
+        long = "storage-nkv-nats-auth-nkey",
+        env = "STORAGE_NATSKV_NATS_AUTH_NKEY"
+    )]
+    storage_nkv_nats_auth_nkey: Option<String>,
```
thomastaylor312, line 124: Nit: It might be easier to have top level nats creds options that get used for _everywhere_ we use nats connections and get overridden by these. I can do this later if you don't want to do this one

bin/main.rs
```diff
@@ -75,6 +102,47 @@ struct Args {
...
+    #[arg(
+        long = "storage-nkv-lattice-bucket-prefix",
+        env = "STORAGE_NATSKV_LATTICE_BUCKET_PREFIX"
+    )]
+    storage_nkv_lattice_bucket_prefix: Option<String>,
+
+    /// (Optional) NATS NKey authentication
+    #[arg(
+        long = "storage-nkv-nats-auth-nkey",
+        env = "STORAGE_NATSKV_NATS_AUTH_NKEY"
```
thomastaylor312, line 122: Nit: Can we shorten this to `WADM_NATS_STORAGE_NKEY`? Nkeys are always an auth thing in context of nats and this feels really long

bin/main.rs
```diff
@@ -75,6 +102,47 @@ struct Args {
...
+    )]
+    storage_nkv_nats_auth_nkey: Option<String>,
+
+    /// (Optional) NATS credential file to use when authenticating
+    #[arg(
+        long = "storage-nkv-nats-auth-creds-file",
+        env = "STORAGE_NATSKV_NATS_AUTH_CREDS_FILE"
+    )]
+    storage_nkv_nats_auth_creds_file: Option<String>,
```
thomastaylor312, line 131: Can you make this mutually exclusive with nkey and seed? It'll make it clear to anyone who tries to specify both

src/storage/mod.rs
thomastaylor312: After looking at the full NATS implementation, I am inclined to agree with @autodidaddict here that it makes things harder to read without adding much value for us at this point in time. I feel like it might be simpler to just do what we need. I offer the following compromise: What if we changed this so it was a single `Store` trait with all the needed methods (still allows for additional implementations if someone needs it) and then had the NATS impl for that trait?

src/storage/state.rs
```diff
@@ -0,0 +1,151 @@
...
+
+use crate::events::Linkdef;
+
+////////////
+// Claims //
+////////////
+
+/// A Claim that can be applied to a given lattice
+#[derive(Debug, Serialize, Deserialize, Clone, Builder, PartialEq, Eq)]
+#[builder(field(public))]
```
thomastaylor312, line 15: Same thing here with the builder stuff on all these structs. It is just easier to derive `Default` and then let people fill out the fields they need

src/storage/state.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/////////////
+// Lattice //
+/////////////
+
+/// Lattice-wide parameters
+#[derive(Debug, Serialize, Deserialize, Clone, Default, PartialEq, Eq)]
+pub struct LatticeParameters {
+    host_status_decay_rate_seconds: Option<u128>,
```
thomastaylor312, line 124: Is this here to match a value we need elsewhere? `u128` is an odd one to see in a public API so I just wanted to check

thomastaylor312 (changes requested):
Good work on the code here. I have some feedback points around some design stuff. I'd also like to note my agreement with @autodidaddict here that we should just have the test pipeline set up NATS for us

.github/workflows/test.yml
```diff
@@ -0,0 +1,44 @@
...
+      # Install: nats-serveer
+      - name: Install nats binary
+        if: ${{ steps.cache-nats.outputs.cache-hit != 'true' }}
+        run: |
+          curl -L https://github.com/nats-io/nats-server/releases/download/v${{ matrix.nats_version }}/nats-server-v${{ matrix.nats_version }}-linux-amd64.zip -o nats-server.zip
+          unzip nats-server.zip -d nats-server
+          sudo cp nats-server/nats-server-v${{ matrix.nats_version }}-linux-amd64/nats-server /usr/bin
+
+      # Run: integration tests for NKV storage
+      - name: Run integration tests [storage-nkv]
```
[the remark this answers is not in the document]
vados-cosmonic: Yeah, they didn't work though so I wasn't trying to take on more scope

bin/main.rs
```diff
@@ -1,4 +1,30 @@
 use std::sync::Arc;
+extern crate derive_builder;
```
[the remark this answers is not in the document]
vados-cosmonic: Yup this is true, I thought I got an error while developing with it, but maybe I'd forgotten to `cargo add` or something, definitely works

src/storage/state.rs
vados-cosmonic: The reason to use Builder here was to make partial updates possible, in particular *before* I realized that's really a good idea with the guarantees NATS KV provides.

Will remove though, in the end the Lattice state has to basically go in all in one go, unfortunately.

src/storage/state.rs
vados-cosmonic: I saw this as a lattice parameter somewhere in some rust code, but will make this a u32!

.github/workflows/test.yml
```diff
@@ -0,0 +1,44 @@
...
+        nats_version: [ 2.9.15 ]
+    steps:
+      - uses: actions/checkout@v2
+
+      # Cache: rust
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: "${{ matrix.os }}-rust-cache"
+
+      # Cache: nats-server binary
```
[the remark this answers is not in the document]
vados-cosmonic: - [ ] add nats service to GH action for integration tests

bin/main.rs
vados-cosmonic: - [ ] make auth creds mutually exclusive with NATS JWT seed

.github/workflows/test.yml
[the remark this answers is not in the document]
vados-cosmonic: fixed this -- it runs the other tests now.

I was able to fix one, but the ones that are failing now are more complex, it looks like it's waiting for something that never happens.

bin/main.rs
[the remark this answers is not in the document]
vados-cosmonic: fixed!

bin/main.rs
vados-cosmonic: done!

src/storage/state.rs
vados-cosmonic: All the Builder stuff was removed!

src/storage/state.rs
vados-cosmonic: This has been changed to `u32`

thomastaylor312:
Those tests ran fine locally before. But we technically had a separate task for the test pipeline. So I'll ignore for now and then come back to the tests

thomastaylor312 (approved):
This works great for now! I'll start using it for my task and clean up any small things that may be left over

vados-cosmonic:
Thanks @thomastaylor312 ! I suspect you'll have to make a few changes as you integrate and actually *use* the thing but glad this was passable :bow:

PR #64, opened 2023-03-17
ref(storage): Refactors store to better fit the data structure we need

thomastaylor312:
This is part of my work for actually generating the lattice state, however I realized that in the previous task we didn't really specify how we were going to save the data. As I started to work on storing lattice state, it became obvious what we had wasn't a good fit. So I refactored to better fit what we needed and took a little bit of time to make the patterns match what we had in other places (such as taking a `Store`, rather than constructing the client and the bucket within the store)

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [X] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [X] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
I updated the integration tests to make sure data could be stored and retrieved

### Manual Verification
Verification was done through integration tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: Test Wadm (v0.4)
 
 on:
   pull_request:
-    branches: [ wadm_0.4 ]
+    branches: [wadm_0.4]
 
 jobs:
   test:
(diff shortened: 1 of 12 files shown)


autodidaddict (approved):
discussed some of the designs herein with @thomastaylor312 . LGTM

PR #65, opened 2023-03-22
feat(*): State generation

thomastaylor312:
This PR adds state generation based off of incoming events. Honestly, I would love for this code to be cleaner, but for now I think this works fine. I've tried to document the code well so people can follow along.

Integration tests will be added in a follow up, but I did test this locally with two wasmcloud hosts and two wadm processes running, while checking the state with `nats kv get`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,8 @@ default = []
 cli = ["clap", "tracing-opentelemetry", "tracing-subscriber", "opentelemetry", "opentelemetry-otlp", "atty", "uuid"]
 
 [dependencies]
+# workaround for exported NATS deps
+alt_nats = {package = "async-nats", version = "0.23"}
 anyhow = "1"
 async-nats = "0.27"
 async-trait = "0.1"
@@ -16,12 +18,14 @@ chrono = "0.4"
 clap = { version = "4", features = ["derive", "cargo", "env"], optional = true }
 cloudevents-sdk = "0.7"
 futures = "0.3"
+nkeys = "0.2.0"
 # One version back to avoid clashes with 0.10 of otlp
 opentelemetry = { version = "0.17", features = ["rt-tokio"], optional = true }
 # 0.10 to avoid protoc dep
 opentelemetry-otlp = { version = "0.10", features = ["http-proto", "reqwest-client"], optional = true }
 # TODO: Actually leverage prometheus
 prometheus = { version = "0.13", optional = true }
+semver = { version = "1.0.16", features = [ "serde" ] }
 serde = "1"
 serde_json = "1"
 sha2 = "0.10.2"
@@ -32,8 +36,7 @@ tracing-futures = "0.2"
 tracing-opentelemetry = { version = "0.17", optional = true }
 tracing-subscriber = { version = "0.3.7", features = ["env-filter", "json"], optional = true }
 uuid = { version = "1", optional = true }
-semver = { version = "1.0.16", features = [ "serde" ] }
-nkeys = "0.2.0"
+wasmcloud-control-interface = "0.24"
 
 [dev-dependencies]
 serial_test = "1"
@@ -42,3 +45,6 @@ serial_test = "1"
 name = "wadm"
 path = "bin/main.rs"
 required-features = ["cli"]
+
+[patch.crates-io]
+wasmcloud-control-interface = { git = "https://github.com/thomastaylor312/control-interface-client.git", branch = "chore/update_rpc_version" }
--- a/bin/main.rs
+++ b/bin/main.rs
@@ -1,10 +1,10 @@
 use std::path::PathBuf;
 use std::sync::Arc;
+use std::time::Duration;
 
-use async_nats::jetstream::Context;
 use clap::Parser;
 use tokio::sync::Semaphore;
-use tracing::{debug, error, trace};
+use tracing::{debug, trace};
 
 use wadm::{
     commands::*,
@@ -12,8 +12,8 @@ use wadm::{
         manager::{ConsumerManager, WorkResult, Worker},
         *,
     },
-    events::*,
-    storage::nats_kv::NatsKvStore,
+    storage::{nats_kv::NatsKvStore, reaper::Reaper},
+    workers::EventWorker,
     DEFAULT_COMMANDS_TOPIC, DEFAULT_EVENTS_TOPIC,
 };
 
@@ -123,6 +123,15 @@ struct Args {
         default_value = "wadm_state"
     )]
     state_bucket: String,
+
+    /// The amount of time in seconds to give for hosts to fail to heartbeat and be removed from the
+    /// store
+    #[arg(
+        long = "cleanup-interval",
+        env = "WADM_CLEANUP_INTERVAL",
+        default_value = "120"
+    )]
+    cleanup_interval: u64,
 }
 
 #[tokio::main]
@@ -136,19 +145,42 @@ async fn main() -> anyhow::Result<()> {
     );
 
     // Build storage adapter for lattice state (on by default)
-    let (_client, context) = nats::get_client_and_context(
+    let (client, context) = nats::get_client_and_context(
+        args.nats_server.clone(),
+        args.domain.clone(),
+        args.nats_seed.clone(),
+        args.nats_jwt.clone(),
+        args.nats_creds.clone(),
+    )
+    .await?;
+
+    let _alt_client = nats::get_alt_client(
         args.nats_server,
-        args.domain,
         args.nats_seed,
         args.nats_jwt,
         args.nats_creds,
     )
     .await?;
 
+    // TODO: We will probably need to set up all the flags (like lattice prefix and topic prefix) down the line
+    let ctl_client_builder = wasmcloud_control_interface::ClientBuilder::new(client);
+    let ctl_client = if let Some(domain) = args.domain {
+        ctl_client_builder
+            .js_domain(domain)
+            .build()
+            .await
+            .map_err(|e| anyhow::anyhow!("{e:?}"))?
+    } else {
+        ctl_client_builder
+            .build()
+            .await
+            .map_err(|e| anyhow::anyhow!("{e:?}"))?
+    };
+
     let store = nats::ensure_kv_bucket(&context, args.state_bucket, 1).await?;
 
     // TODO: use the storage engine
-    let _state_storage = NatsKvStore::new(store);
+    let state_storage = NatsKvStore::new(store);
 
     let event_stream = nats::ensure_stream(
         &context,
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -42,3 +45,6 @@ serial_test = "1"
 name = "wadm"
 path = "bin/main.rs"
 required-features = ["cli"]
+
+[patch.crates-io]
+wasmcloud-control-interface = { git = "https://github.com/thomastaylor312/control-interface-client.git", branch = "chore/update_rpc_version" }
```
thomastaylor312, line 50: I'll change this once https://github.com/wasmCloud/control-interface-client/pull/44 is merged

autodidaddict (commented):
I don't know if you're planning on doing this for followup, but what I'd love to see here is a set of unit-tested logic (similar to the lattice observer in Elixir) where you can assert that a given set of events produces a given state ... which would then decouple that state somewhat from the underlying KV store. Right now it reads like the KV store is tightly coupled to the 'business' logic of calculating state.

autodidaddict:
Approving for now but as mentioned in the preceding commend, I'd like to see the logic that calculates state to be something that is testable in isolation.

autodidaddict (approved):

src/workers/event.rs
```diff
@@ -0,0 +1,545 @@
...
+        trace!("Storing updated actors in store");
+        self.store.store_many(lattice_id, actors_to_update).await?;
+
+        trace!("Fetching providers from store to remove stopped instances");
+        let mut all_providers = self.store.list::<Provider>(lattice_id).await?;
+
+        let providers_to_update = current.providers.into_iter().filter_map(|info| {
+            let key = crate::storage::provider_id(&info.public_key, &info.link_name);
+            let mut prov = all_providers.remove(&key)?;
+            prov.hosts.remove(&host.id).map(|_| (key, prov))
```
connorsmith256, line 197: I found this logic to be confusing, and I'm still not sure I grok it. Here's what I think we're doing:
- retrieve `all_providers` running in the lattice
- for each provider that was running on `current`, remove it from `all_providers`
  - question: why `remove` instead of `get` or `entry`? A provider with a unique `(pk, link)` ID can exist multiple times within the lattice (once per host). Calling `remove` caused me to infer we were purposefully removing from the lattice state _entirely_, which would be incorrect; but we're _not_ actually doing that, since we're not referring to `all_providers` after this loop, and since we're only processing one host ID here. So this seems okay, but unnecessarily confusing and prone to future bugs. I assume you did this to get a `Provider` instead of a `&Provider`, so you could call `prov.hosts.remove`. IMO this is an example where Rust's type system is incentivizing confusing code and I'd rather work around the types to write clearer code
- if we found a provider with the given key, try to remove the stopped host from its map. If we get a result, return a `(key, prov)` tuple as the new state we'll store. Otherwise return None since this entry doesn't need to be updated

Assuming I've understood this, I think something like this is more clear:
```rust
let mut maybe_prov = (*all_providers.get(&key)).clone();
match maybe_prov {
  Some(prov) => match prov.hosts.remove(&host.id) {
    Some(_) => (key, prov)
    None => None // provider wasn't running on this host
  }
  None => None // provider wasn't running in the lattice (maybe warn here? This seems unexpected)
}
```

Normally I prefer `map` to matching like this, but given the complexity here it might be worth calling out the individual cases

src/workers/event.rs
```diff
@@ -0,0 +1,545 @@
...
+        let mut actors = self.store.list::<Actor>(lattice_id).await?;
+
+        let missing_actors: HashSet<String> = host
+            .actors
+            .iter()
+            .filter_map(|(id, _)| (!actors.contains_key(id)).then_some(id))
+            .cloned()
+            .collect();
+
+        // TODO: FIXME
```
connorsmith256, line 397: What's this? Looks like maybe there's some leftover duplicate code here?

src/workers/event.rs
```diff
@@ -0,0 +1,545 @@
...
+                        None
+                    } else {
+                        Some((id.to_owned(), current))
+                    }
+                }
+                None => None,
+            }
+        });
+
+        let actors_to_update: Vec<(String, Actor)> = if !missing_actors.is_empty() {
```
connorsmith256, line 418: So `missing_actors` covers the case where the lattice state we have doesn't include some actors on the heartbeat. But what about the reverse case, where the lattice state includes actors that aren't on the heartbeat? E.g. if we missed an ActorStopped event

src/workers/event.rs
```diff
@@ -0,0 +1,545 @@
...
+    async fn heartbeat_provider_update(
+        &self,
+        lattice_id: &str,
+        host: &HostHeartbeat,
+    ) -> anyhow::Result<()> {
+        debug!("Fetching current provider state");
+        let mut providers = self.store.list::<Provider>(lattice_id).await?;
+        let providers_to_update = host.providers.iter().filter_map(|info| {
+            let provider_id = crate::storage::provider_id(&info.public_key, &info.link_name);
+            match providers.remove(&provider_id) {
```
connorsmith256, line 447: Same comment here as above, I find the `remove` quite confusing

connorsmith256 (changes requested):

src/consumers/events.rs
```diff
@@ -113,7 +113,7 @@ impl Stream for EventConsumer {
                 let evt = match Event::try_from(raw_evt) {
                     Ok(evt) => evt,
                     Err(e) => {
-                        warn!(error = ?e, "Unable to decode as event. Skipping message");
+                        debug!(error = ?e, "Unable to decode as event. Skipping message");
```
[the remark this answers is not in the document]
thomastaylor312: Because there are several events that we purposefully aren't doing anything with yet (like `refmap_set`)

bin/main.rs
```diff
@@ -136,19 +145,42 @@ async fn main() -> anyhow::Result<()> {
...
+    let (client, context) = nats::get_client_and_context(
+        args.nats_server.clone(),
+        args.domain.clone(),
+        args.nats_seed.clone(),
+        args.nats_jwt.clone(),
+        args.nats_creds.clone(),
+    )
+    .await?;
+
+    let _alt_client = nats::get_alt_client(
```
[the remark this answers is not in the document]
thomastaylor312: This was to get around the mismatched nats version that I thought I was going to have to deal with. But now that I updated it, I can remove this

src/storage/reaper.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+}
+
+#[instrument(level = "debug", skip(store))]
+async fn reaper_fn<S: Store>(store: S, lattice_id: String, check_interval: Duration) {
+    debug!("Starting reaper");
+    // SAFETY: We created this Duration from a std Duration, so it should unwrap back just fine
+    let mut ticker = time::interval(check_interval.to_std().unwrap());
+    ticker.set_missed_tick_behavior(time::MissedTickBehavior::Delay);
```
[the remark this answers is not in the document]
thomastaylor312: In the odd case we get a missed tick, I don't want to end up triggering two close ticks (along with associated database reads and writes). This just basically picks us back up on the same interval as before. Honestly it probably doesn't matter _too_ much if you think it is confusing. But was just trying to at least think about it a little

src/workers/event.rs
```diff
@@ -0,0 +1,545 @@
...
+            trace!(actor = ?current, "Found existing actor data");
+            // Merge in current counts
+            actor_data.count = current.count;
+        }
+
+        // Update count of the data
+        actor_data
+            .count
+            .entry(actor.host_id.clone())
+            .and_modify(|val| *val += 1)
```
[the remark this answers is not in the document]
thomastaylor312: Oh yep good catch. I refactored this towards the end and missed that. I think the better option here is just to default the count map to nothing. I'll also try to match the behavior with provider

src/workers/event.rs
```diff
@@ -0,0 +1,545 @@
...
+                    trace!(host_id = %actor.host_id, "Stopped last actor on host, removing host entry from actor");
+                    current.count.remove(&actor.host_id);
+                } else {
+                    trace!(count = %new_count, host_id = %actor.host_id, "Setting current actor");
+                    current.count.insert(actor.host_id.clone(), new_count);
+                }
+            }
+
+            if current.count.is_empty() {
+                debug!("Last actor instance was removed, removing actor from storage");
```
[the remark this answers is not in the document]
thomastaylor312: Oh yep, probably changed one of them and then forgot to change the other

src/workers/event.rs
```diff
@@ -0,0 +1,545 @@
...
+        debug!("Updating store with current host heartbeat information");
+        // Host updates just overwrite current information, so no need to fetch
+        let host_data = Host::from(host);
+        self.store
+            .store(lattice_id, host.id.clone(), host_data)
+            .await?;
+
+        // NOTE: We can return an error here and then nack because we'll just reupdate the host data
+        // with the exact same host heartbeat entry. There is no possibility of a duplicate
+        self.heartbeat_provider_update(lattice_id, host).await?;
```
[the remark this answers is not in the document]
thomastaylor312: Nope, it just results in a redelivery. Technically any of the events can do this, but my comment here was noting that we won't end up with duplicate counts by returning an error here (because we are just overwriting data). By default, it takes 3 message redeliveries to have NATS poison pill us

src/workers/event.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    #[instrument(level = "debug", skip(self, host), fields(host_id = %host.id))]
+    async fn handle_host_started(
+        &self,
+        lattice_id: &str,
+        host: &HostStarted,
+    ) -> anyhow::Result<()> {
+        debug!("Updating store with new host");
+        // New hosts have nothing running on them yet, so just drop it in the store
```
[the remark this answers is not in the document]
thomastaylor312: I _think_ those still trigger the proper actor/provider started events? @autodidaddict can you confirm?

src/workers/event.rs
thomastaylor312: Yep 🤣. I discovered this yesterday when I went to add a comment

src/workers/event.rs
thomastaylor312: Yes, you read the logic properly and I was trying to avoid the clone (which is why I was _removing_ from the `all_providers` map, which is owned data). We're probably just fine to clone here and I'll make a note that it was a purposeful tradeoff

src/workers/event.rs
thomastaylor312: Yep 🤣. I discovered this yesterday when I went to add a comment

src/workers/event.rs
thomastaylor312: Yes, you read the logic properly and I was trying to avoid the clone (which is why I was _removing_ from the `all_providers` map, which is owned data). We're probably just fine to clone here and I'll make a note that it was a purposeful tradeoff

thomastaylor312: Just a heads up that clippy didn't like that inner match, so I stuck with the `remove().map()`

src/workers/event.rs
thomastaylor312: I addressed this with some stuff I added to the reaper process

src/storage/reaper.rs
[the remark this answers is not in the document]
connorsmith256: Gotcha. I think either is fine. I doubt we'll ever miss a tick in practice, since our ticks are so long

connorsmith256 (approved):

Issue #43, opened 2023-03-04
Receive and parse wadm/OAM manifests


thomastaylor312:
To be complete, there should be a module in the library code that:

*   Has concrete types for wadm manifests that can deserialize and serialize to yaml and json
*   Listens for manifests on the correct topics (see existing wadm code)
*   Properly parses those manifests

thomastaylor312:
After talking to @stephenandary, I am going to start work on this for now

thomastaylor312:
Closed with #72 and #73

PR #66, opened 2023-03-24
Added parser to serialize and deserialize OAM manifests

aish-where-ya:
## Feature or Problem
Parses OAM manifests for warm. Supports YAML and JSON configurations and converts them into in memory Rust structs.

## Related Issues
#43 

## Release Information
next

## Consumer Impact
Will support parsing option for manifests

## Testing
Unit tests added

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
test_oam_deserializer
test_oam_serializer

### Acceptance or Integration
NA

### Manual Verification
Manually verified with existing YAML/JSON files.

### DIFF
--- a/src/events/parser.rs
+++ b/src/events/parser.rs
@@ -1,28 +1,27 @@
-use anyhow::Result;
-use serde_json;
-use serde_yaml;
 use std::collections::BTreeMap;
-use std::io::BufReader;
-use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
 
+pub const DEFAULT_WEIGHT: i32 = 100;
+
 #[derive(Debug, Serialize, Deserialize)]
 pub enum ComponentType {
-    actor,
-    capability,
+    #[serde(rename = "actor")]
+    Actor,
+    #[serde(rename = "capability")]
+    Capability,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Annotations {
-    version: String,
-    description: String,
+    pub version: String,
+    pub description: String,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Metadata {
-    name: String,
-    annotations: Annotations,
+    pub name: String,
+    pub annotations: Annotations,
 }
 
 fn default_link_name() -> Option<String> {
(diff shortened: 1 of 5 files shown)


src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+use serde_yaml;
+use std::collections::BTreeMap;
+use std::io::BufReader;
+use std::path::PathBuf;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum ComponentType {
+    actor,
```
thomastaylor312, line 12: Rust enums should have uppercase names. If you need to match something from deserialization, you can use the `rename` attribute: https://serde.rs/variant-attrs.html#rename

src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+fn default_link_name() -> Option<String> {
+    Some("default".to_string())
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Properties {
+    image: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    contract: Option<String>,
+    #[serde(default = "default_link_name", skip_serializing_if = "Option::is_none")]
```
thomastaylor312, line 37: For this `Option` and on line 35, you shouldn't need a default since it is an `Option`

src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+    link_name: Option<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum TraitType {
+    spreadscaler,
```
thomastaylor312, line 43: Same thing here will all the enums

src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+    linkdef,
+}
+#[derive(Debug, Serialize, Deserialize)]
+pub enum SpreadRequirements {
```
thomastaylor312, line 47: So the `requirements` object is actually just a map of string, string because they can be entirely arbitrary. So all you should need for `requirements` is `BTreeMap<String, String>` rather than `BTreeMap<String, SpreadRequirementValues>`

src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+    app(String),
+    zone(String),
+    ledenabled(bool),
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum LinkdefValueItems {
```
thomastaylor312, line 54: Based on the OAM model we have, this should be a struct with members named the same as you have below

src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+    uri(String),
+    address(String),
+    port(i32),
+}
+
+fn default_weight() -> Option<i32> {
```
thomastaylor312, line 60: Nit: Rather than set the option when deserializing, it is a bit more idiomatic in Rust to create a const like `DEFAULT_WEIGHT` and then when we are accessing the data we can do `spread.weight.unwrap_or(DEFAULT_WEIGHT)`

src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Manifest {
+    apiVersion: String,
+    kind: String,
+    metadata: Metadata,
+    spec: Components,
+}
+
+pub(crate) fn deserialize_yaml(filepath: PathBuf) -> Result<Manifest> {
```
thomastaylor312, line 123: These probably belong in a test as we will only be reading them from NATS and the KV store when using them

src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+    Ok(json_string)
+}
+
+#[cfg(test)]
+mod test {
+
+    use super::*;
+
+    #[test]
+    fn test_oam_desrializer() {
```
thomastaylor312, line 145: Spelling nit: `deserializer`

src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+    name: String,
+    requirements: BTreeMap<String, SpreadRequirementValues>,
+    #[serde(default = "default_weight", skip_serializing_if = "Option::is_none")]
+    weight: Option<i32>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum TraitPropertyItem {
+    LinkdefProperty {
```
thomastaylor312, line 84: Could you make each of these be their own structs and then include them here? e.g.

```rust
pub struct SpreadScalerProperty {
        replicas: i32,
        spread: Vec<Spread>,
    }
    
 pub enum TraitPropertyItem {
     SpreadScaler(SpreadScalerProperty)
 }
 ```
 
 The reason being that we will need to use those structs in our future `Scaler` types as config data

File: src/events/parser.rs [hunk 10]
```diff
@@ -0,0 +1,280 @@
...
+    properties: TraitPropertyItem,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ComponentItem {
+    name: String,
+    #[serde(rename = "type")]
+    component_type: ComponentType,
+    properties: Properties,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
thomastaylor312, line 107: No default needed here

src/events/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum ComponentType {
+    actor,
+    capability,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Annotations {
+    version: String,
```
thomastaylor312, line 18: Can you please make all the data members for all the structs in here `pub`? In rust you have to label each field as public (e.g. `pub version: String`)

thomastaylor312 (changes requested):

src/events/parser.rs
aish-where-ya: Putting the value as String fails because if you checkout [this](https://github.com/wasmCloud/wadm/blob/main/oam/simple1.yaml) manifest on line 53, the value is treated as a boolean instead of String. Which is why the Map needs to accept a variety of values.

thomastaylor312: I think this is a case where we can break with what wadm is currently doing. A map of string, string is what host labels are and that is what we are currently matching. Plus it makes the code much easier. So I think it'd be ok to change

thomastaylor312 (approved):
This is decently close, so going to merge for now and finish up the last remaining comment in the next PR

PR #69, opened 2023-03-30
feat(event): Full state unit tests

thomastaylor312:
This adds a lot of unit tests to make sure state is generated properly. They are quite long but they walk through a roughly lifelike flow of events

### DIFF
--- a/src/test_util.rs
+++ b/src/test_util.rs
@@ -1,0 +1,96 @@
+use std::collections::HashMap;
+use std::convert::Infallible;
+
+use serde::{de::DeserializeOwned, Serialize};
+
+use crate::storage::StateKind;
+
+fn generate_key<T: StateKind>(lattice_id: &str) -> String {
+    format!("{}_{lattice_id}", T::KIND)
+}
+
+/// A [`Store`] implementation for use in testing
+#[derive(Default)]
+pub struct TestStore {
+    pub inner: tokio::sync::RwLock<HashMap<String, Vec<u8>>>,
+}
+
+#[async_trait::async_trait]
+impl crate::storage::Store for TestStore {
+    type Error = Infallible;
+
+    async fn get<T>(&self, lattice_id: &str, id: &str) -> Result<Option<T>, Self::Error>
+    where
+        T: DeserializeOwned + StateKind,
+    {
+        let key = generate_key::<T>(lattice_id);
+        let mut all: HashMap<String, T> = self
+            .inner
+            .read()
+            .await
+            .get(&key)
+            .map(|raw| serde_json::from_slice(raw).unwrap())
+            .unwrap_or_default();
+        // T isn't clone, so I can't use get
+        Ok(all.remove(id))
+    }
+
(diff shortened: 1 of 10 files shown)


src/test_util.rs
```diff
@@ -0,0 +1,96 @@
...
+        let key = generate_key::<T>(lattice_id);
+        let mut all: HashMap<String, T> = self
+            .inner
+            .read()
+            .await
+            .get(&key)
+            .map(|raw| serde_json::from_slice(raw).unwrap())
+            .unwrap_or_default();
+        // T isn't clone, so I can't use get
+        Ok(all.remove(id))
```
brooksmtownsend, line 35: Does this mean we for-realz remove it when we do this call? Or is it a copy

src/workers/event.rs
```diff
@@ -355,28 +428,18 @@ impl<S: Store + Send + Sync> EventWorker<S> {
...
                             id: id.clone(),
-                            name: claim.remove("name").unwrap_or_default(),
-                            capabilities: claim
-                                .remove("caps")
-                                .map(|raw| raw.split(',').map(|s| s.to_owned()).collect())
-                                .unwrap_or_default(),
-                            issuer: claim.remove("iss").unwrap_or_default(),
+                            name: claim.name,
+                            capabilities: claim.capabilities,
+                            issuer: claim.issuer,
```
brooksmtownsend, line 442: Much cleaner 👌🏻

src/workers/event.rs
```diff
@@ -556,3 +619,892 @@ impl<S: Store + Send + Sync> Worker for EventWorker<S> {
...
+        let host2_id = "starkiller".to_string();
+
+        /***********************************************************/
+        /******************** Host Start Tests *********************/
+        /***********************************************************/
+        let labels = HashMap::from([("superweapon".to_string(), "true".to_string())]);
+        worker
+            .handle_host_started(
+                lattice_id,
+                &HostStarted {
```
brooksmtownsend, line 657: I think if we write more tests, it would be helpful to have simple functions for generating this host started event. E.g.
```
fn host_started_evt("spongebobid", "sPoNgEbObId", [("sponge".to_string(), "bob".to_string())])
```

May not end up saving too much but it's nice

brooksmtownsend (approved):
This all looks pretty wonderful, albeit many lines of code. Left a comment that we may want to consider a simpler way of generating this state, whether that's some helper functions (which would probably be just as much work as the raw struct with the HostHeartbeat) or maybe some format that we can store in JSON and create state from. Nothing that needs to hold this back because I don't have a full concrete proposal to make it better :)

src/test_util.rs
thomastaylor312: It is just a copy as we are reading a HashMap from our in memory HashMap (yo dawg)

thomastaylor312:
Yeah it could save a few lines of code, but I am also very done with writing these tests 😆 . So I'll hold out for now

PR #70, opened 2023-03-30
ref(manifests): Clearer manifest types and documentation

thomastaylor312:
This moves the manifests to their own `mod` with additional documentation. I also fixed some issues with places where we needed a map instead of a concrete struct


src/model.rs
```diff
@@ -2,129 +2,172 @@ use std::collections::BTreeMap;
...
 }
 
+/// Properties for linkdefs
 #[derive(Debug, Serialize, Deserialize)]
 pub struct LinkdefProperty {
+    /// The target this linkdef applies to. This should be the name of an actor component
     pub target: String,
-    pub values: LinkdefValueItems,
+    /// Values to use for this linkdef
+    pub values: BTreeMap<String, String>,
```
brooksmtownsend, line 136: nit: we should skip serializing this if empty

src/model.rs
```diff
@@ -2,129 +2,172 @@ use std::collections::BTreeMap;
...
-#[derive(Debug, Serialize, Deserialize)]
-pub struct ComponentItem {
+pub struct Spread {
+    /// The name of this spread requirement
     pub name: String,
-    #[serde(rename = "type")]
-    pub component_type: ComponentType,
-    pub properties: Properties,
+    /// An arbitrary map of labels to match on for scaling requirements
+    pub requirements: BTreeMap<String, String>,
```
brooksmtownsend, line 154: nit: skip if empty

src/model.rs
```diff
@@ -2,129 +2,172 @@ use std::collections::BTreeMap;
...
-}
-
-#[derive(Debug, Serialize, Deserialize)]
-pub struct Manifest {
-    #[serde(rename = "apiVersion")]
-    pub api_version: String,
-    pub kind: String,
-    pub metadata: Metadata,
-    pub spec: Components,
+    pub weight: Option<usize>,
```
brooksmtownsend, line 157: nit: skip if none

brooksmtownsend (approved):
Looks great 😄 I left some nits around serde serializing, but I'm really not sure if they have an effect so I'm approving

src/model.rs
thomastaylor312: This is already there, the diff just makes it looks like it hasn't

Issue #48, opened 2023-03-04
Spike: Define `Scaler` trait


thomastaylor312:
A scaler will need to take a config of some kind, likely a handle to some sort of state, and two different methods. Below is a general idea of what it could look like, but part of this spike is to have an integration test that can pass in an `Event` or a wadm manifest and output commands with a basic, not fully functional spreadscaler.

```rust
/// A Scaler is used to manage responding to even
pub trait Scaler {
  /// Any type that has the necessary data to configure the scaler
  type Config: Send + Sync;

  /// Handles the event, returning any needed changes in response
  fn handle_event(&self, event: ScopedEvent) -> Result<HashSet<Command>>;

  /// Handles a new or updated manifest with its given config. This configuration should be stored by implementors in some form so that handle_event can produce the right commands
  fn handle_manifest(&self, config: Self::Config) -> Result<HashSet<Command>>;

  /// Removes a config from the scaler and emits the expected compensatory commands
  fn remove_manifest(&self, config: &Self::Config) -> Result<HashSet<Command>>;
}
```

thomastaylor312:
Closed in #71

PR #71, opened 2023-03-31
Scaler trait and sample implementation with a SimpleScaler

brooksmtownsend:
## Primary Reviewer
@thomastaylor312 

The Scaler trait, which defines a trait that can be implemented to handle the scaling of a single component (be it actors, linkdefs, or providers) looks like this:
```rust
#[async_trait]
pub trait Scaler {
    type Config: Send + Sync;

    /// Provide a scaler with configuration to use internally when computing commands
    fn update_config(&mut self, config: Self::Config) -> Result<bool>;

    /// Compute commands that must be taken given an event that changes the lattice state
    async fn handle_event<S: ReadStore + Send + Sync>(
        &self,
        store: S,
        event: Event,
    ) -> Result<HashSet<Command>>;

    /// Compute commands that must be taken to achieve desired state as specified in config
    async fn reconcile<S: ReadStore + Send + Sync>(&self, store: S) -> Result<HashSet<Command>>;
}
```

Essentially, you construct a scaler and give it configuration to use for events, and then it has a function you can call at any point to reconcile between current and desired state. Each event coming in that updates the state should also be passed to the scaler, which can then trigger compensating commands to return to desired state. The `reconcile` function can also be used as a helper function to determine if the Scaler is currently satisfied (e.g. `scaler.reconcile(store).unwrap().is_empty()`)

This PR includes a sample implementation, `SimpleActorScaler`, that uses a set config to compute required start/stop actor commands to reach a number of replicas. If it would better illustrate the flexibility of the trait, I could also add in a `SimpleProviderScaler` or `SimpleLinkdefScaler` which monitors events, but I felt the actor implementation would be enough to illustrate.

## Feature or Problem
Addresses #48 

## Related Issues
#48 

## Release Information
wadm_0.4

## Consumer Impact
Consumers of wadm can implement their own scalers with custom config, specified in a manifest. 

This PR also separates the `Store` trait into a `ReadStore` and a supertrait `Store: ReadStore`. This allows a single Store implementation to easily give read-only access to a store to different code paths, which is useful to ensure that state is not modified when it only needs to be queried.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Wrote two basic unit tests to ensure the SimpleActorScaler can handle enforcing a number of actor replicas on a lattice

### Acceptance or Integration
Unit tests

### Manual Verification
Creating the simplescaler is partial manual verification that the API is ergonomic and works well to describe what kind of thing we are trying to assert with a Scaler. I ensured that I wrote a test to validate that the `Hash` implementation does not block us from returning, say, two StopActor commands with different parameters.

### DIFF
--- a/src/scaler/mod.rs
+++ b/src/scaler/mod.rs
@@ -2,15 +2,16 @@ use anyhow::Result;
 use async_trait::async_trait;
 use std::collections::HashSet;
 
-use crate::{commands::Command, events::Event, storage::ReadStore};
+use crate::{commands::Command, events::Event};
 
 mod simplescaler;
 
 /// A trait describing a struct that can be configured to compute the difference between
 /// desired state and configured state, returning a set of commands to approach desired state.
 ///
-/// State is given to a scaler using a [ReadStore](crate::storage::ReadStore) so that it can retrieve
-/// current information from the proper store without changing state directly.
+/// Implementers of this trait can choose how to access state, but it's generally recommended to
+/// use a [ReadStore](crate::storage::ReadStore) so that it can retrieve current information about
+/// state using a common trait that only allows store access and not modification
 ///
 /// Typically a Scaler should be configured with `update_config`, then use the `reconcile` method
 /// for an inital set of commands. As events change the state, they should also be given to the Scaler
@@ -23,12 +24,8 @@ pub trait Scaler {
     fn update_config(&mut self, config: Self::Config) -> Result<bool>;
 
     /// Compute commands that must be taken given an event that changes the lattice state
-    async fn handle_event<S: ReadStore + Send + Sync>(
-        &self,
-        store: S,
-        event: Event,
-    ) -> Result<HashSet<Command>>;
+    async fn handle_event(&self, event: Event) -> Result<HashSet<Command>>;
 
     /// Compute commands that must be taken to achieve desired state as specified in config
-    async fn reconcile<S: ReadStore + Send + Sync>(&self, store: S) -> Result<HashSet<Command>>;
+    async fn reconcile(&self) -> Result<HashSet<Command>>;
 }
--- a/src/scaler/mod.rs
+++ b/src/scaler/mod.rs
@@ -1,0 +1,34 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use std::collections::HashSet;
+
+use crate::{commands::Command, events::Event, storage::ReadStore};
+
+mod simplescaler;
+
+/// A trait describing a struct that can be configured to compute the difference between
+/// desired state and configured state, returning a set of commands to approach desired state.
+///
+/// State is given to a scaler using a [ReadStore](crate::storage::ReadStore) so that it can retrieve
+/// current information from the proper store without changing state directly.
+///
+/// Typically a Scaler should be configured with `update_config`, then use the `reconcile` method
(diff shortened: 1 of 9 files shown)


src/scaler/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+pub trait Scaler {
+    type Config: Send + Sync;
+
+    /// Provide a scaler with configuration to use internally when computing commands
+    fn update_config(&mut self, config: Self::Config) -> Result<bool>;
+
+    /// Compute commands that must be taken given an event that changes the lattice state
+    async fn handle_event<S: ReadStore + Send + Sync>(
+        &self,
+        store: S,
```
thomastaylor312, line 28: I think the `store` should just be an implementation detail of the struct that implements `Scaler`. An example of this: we will have a linkdef scaler, which will not be using a `Store` as it will be using a control client to get the linkdefs

src/scaler/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+        event: Event,
+    ) -> Result<HashSet<Command>>;
+
+    /// Compute commands that must be taken to achieve desired state as specified in config
+    async fn reconcile<S: ReadStore + Send + Sync>(&self, store: S) -> Result<HashSet<Command>>;
```
thomastaylor312, line 33: I am not entirely sure if we'll need this yet (we don't have a synthetic "resync" event right now), but I think this works for now

src/scaler/simplescaler.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        let store = Arc::new(TestStore::default());
+        let worker = EventWorker::new(store.clone(), HashMap::default());
+
+        let host_id = "NASDASDIAMAREALHOST".to_string();
+        let host_name = "I am a real host".to_string();
+
+        let labels = HashMap::from([("real".to_string(), "true".to_string())]);
+
+        worker
```
thomastaylor312, line 218: Nit: You could make your life easier if you want and just do `store.store(lattice_id, Host{})`. It is also probably better than just setting the acker to `None` 😆

thomastaylor312 (changes requested):
Just one point of discussion around the actual trait. Otherwise this is looking pretty good

src/scaler/mod.rs
brooksmtownsend: I was mostly thinking of this for the first time you ever reconcile, so you can kick off the required commands for the first time. Could likely accomplish similar effects with a synthetic event

src/scaler/mod.rs
brooksmtownsend: That sounds reasonable and I don't think it'll affect my testing strategy

thomastaylor312 (approved):

src/scaler/simplescaler.rs
brooksmtownsend: Oh, I just ended up not worrying about it. But will do this from now on (if I think about it)!

PR #72, opened 2023-04-01
feat(api): Implements all model operations

### LINKED ISSUE #43: Receive and parse wadm/OAM manifests

### LINKED ISSUE #44: Store wadm manifests in a KV bucket
Received manifests should be stored in a KV bucket inside of NATS:

*   Bucket should auto create if it doesn't exist
*   There should be a struct in code that can perform all CRUD operations for manifests, this involves storing under the right key and scoped to the proper lattice ID
*   Data should be serialized to json for storage

thomastaylor312:
This PR implements the wadm NATS api and backend storage for the manifests. Anywhere I have deviated from the current wadm_0.3 api, I have tried to document it with a note so we can update the docs later. In a follow up PR, I will finish implementing the actual deployment operations of the API. For now, I wanted to make sure this didn't get _too_ big to review.

Partially closes #43 and closes #44

### DIFF
--- a/bin/main.rs
+++ b/bin/main.rs
@@ -12,6 +12,7 @@ use wadm::{
         manager::{ConsumerManager, WorkResult, Worker},
         *,
     },
+    server::{Server, DEFAULT_WADM_TOPIC_PREFIX},
     storage::{nats_kv::NatsKvStore, reaper::Reaper},
     workers::EventWorker,
     DEFAULT_COMMANDS_TOPIC, DEFAULT_EVENTS_TOPIC,
@@ -132,6 +133,23 @@ struct Args {
         default_value = "120"
     )]
     cleanup_interval: u64,
+
+    /// The API topic prefix to use. This is an advanced setting that should only be used if you
+    /// know what you are doing
+    #[arg(
+        long = "api-prefix",
+        env = "WADM_API_PREFIX",
+        default_value = DEFAULT_WADM_TOPIC_PREFIX
+    )]
+    api_prefix: String,
+
+    /// Name of the bucket used for storage of manifests
+    #[arg(
+        long = "manifest-bucket-name",
+        env = "WADM_MANIFEST_BUCKET_NAME",
+        default_value = "wadm_manifests"
+    )]
+    manifest_bucket: String,
 }
 
 #[tokio::main]
@@ -155,7 +173,7 @@ async fn main() -> anyhow::Result<()> {
     .await?;
 
     // TODO: We will probably need to set up all the flags (like lattice prefix and topic prefix) down the line
-    let ctl_client_builder = wasmcloud_control_interface::ClientBuilder::new(client);
+    let ctl_client_builder = wasmcloud_control_interface::ClientBuilder::new(client.clone());
     let ctl_client = if let Some(domain) = args.domain {
         ctl_client_builder
             .js_domain(domain)
@@ -171,8 +189,11 @@ async fn main() -> anyhow::Result<()> {
 
     let store = nats::ensure_kv_bucket(&context, args.state_bucket, 1).await?;
 
-    // TODO: use the storage engine
     let state_storage = NatsKvStore::new(store);
+
+    let store = nats::ensure_kv_bucket(&context, args.manifest_bucket, 1).await?;
+
+    let manifest_storage = NatsKvStore::new(store);
 
     let event_stream = nats::ensure_stream(
         &context,
@@ -222,7 +243,14 @@ async fn main() -> anyhow::Result<()> {
         Duration::from_secs(args.cleanup_interval / 2),
         ["default".to_owned()],
     );
-    tokio::signal::ctrl_c().await?;
+
+    let server = Server::new(manifest_storage, client, Some(&args.api_prefix)).await?;
+    tokio::select! {
+        res = server.serve() => {
+            res?
+        }
+        _ = tokio::signal::ctrl_c() => {}
+    }
     Ok(())
 }
 
--- a/src/model/internal.rs
+++ b/src/model/internal.rs
@@ -5,6 +5,8 @@ use serde::{Deserialize, Serialize};
 use crate::model::Manifest;
 use crate::storage::StateKind;
 
+/// This struct represents a single manfiest, with its version history. Internally these are stored
+/// as an indexmap keyed by version name
 #[derive(Debug, Serialize, Deserialize, Default)]
 pub(crate) struct StoredManifest {
     // Ordering matters for how we store a manifest, so we need to use an index map to preserve
@@ -94,7 +96,7 @@ impl StoredManifest {
     }
 
     /// Returns whether or not this is a new (empty) manifest
-    pub fn is_new(&self) -> bool {
+    pub fn is_empty(&self) -> bool {
         self.manifests.is_empty()
     }
 
--- a/src/model/internal.rs
+++ b/src/model/internal.rs
@@ -1,0 +1,105 @@
+//! Contains the internal storage definition of a manifest
+use indexmap::IndexMap;
+use serde::{Deserialize, Serialize};
+
+use crate::model::Manifest;
+use crate::storage::StateKind;
+
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub(crate) struct StoredManifest {
+    // Ordering matters for how we store a manifest, so we need to use an index map to preserve
+    // insertion order _and_ have quick access to specific versions
+    manifests: IndexMap<String, Manifest>,
+    // Set only if a version is deployed
+    deployed_version: Option<String>,
+    // TODO: Figure out which status to store (probably just the status for each component). We can
+    // convert it to the external facing type from the server as needed
+}
+
+impl StateKind for StoredManifest {
+    const KIND: &'static str = "manifest";
+}
+
+impl StoredManifest {
+    /// Gets the current version of the manifest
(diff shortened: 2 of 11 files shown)


thomastaylor312:
Please note that I am going to add in some integration tests in a separate commit, but the PR is pretty much ready for review otherwise. It is marked as draft purely to make sure we don't merge before I do

bin/main.rs
```diff
@@ -132,6 +133,23 @@ struct Args {
...
     cleanup_interval: u64,
+
+    /// The API topic prefix to use. This is an advanced setting that should only be used if you
+    /// know what you are doing
+    #[arg(
+        long = "api-prefix",
+        env = "WADM_API_PREFIX",
+        default_value = DEFAULT_WADM_TOPIC_PREFIX
+    )]
+    api_prefix: String,
```
connorsmith256, line 144: When would this and the below bucket option be overridden?

bin/main.rs
```diff
@@ -171,9 +189,12 @@ async fn main() -> anyhow::Result<()> {
 
     let store = nats::ensure_kv_bucket(&context, args.state_bucket, 1).await?;
 
-    // TODO: use the storage engine
     let state_storage = NatsKvStore::new(store);
 
+    let store = nats::ensure_kv_bucket(&context, args.manifest_bucket, 1).await?;
```
connorsmith256, line 194: Not sure if this diff is weird but it looks like this line is duplicated

src/model/mod.rs
```diff
@@ -27,6 +35,25 @@ pub struct Manifest {
...
 }
 
+impl Manifest {
+    /// Returns a reference to the current version
+    pub fn version(&self) -> &str {
+        self.metadata
+            .annotations
+            .get(VERSION_ANNOTATION_KEY)
+            .map(|v| v.as_str())
+            .unwrap_or_default()
```
connorsmith256, line 45: Is an empty string okay here? Seems like this might need to generate an error

src/model/internal.rs
```diff
@@ -0,0 +1,105 @@
+//! Contains the internal storage definition of a manifest
+use indexmap::IndexMap;
+use serde::{Deserialize, Serialize};
+
+use crate::model::Manifest;
+use crate::storage::StateKind;
+
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub(crate) struct StoredManifest {
```
connorsmith256, line 9: Potential nit: this struct represents a single manifest with all of its version history, but that wasn't immediately clear. I thought it might be storage of all current manifests until I kept reading. A comment or different type for the key into the index map would help

src/model/internal.rs
```diff
@@ -0,0 +1,105 @@
...
+            .expect("A manifest should always exist. This is programmer error")
+    }
+
+    /// Gets a reference to the specified version of the manifest
+    pub fn get_version(&self, version: &str) -> Option<&Manifest> {
+        self.manifests.get(version)
+    }
+
+    /// Returns whether or not this is a new (empty) manifest
+    pub fn is_new(&self) -> bool {
```
connorsmith256, line 97: Is there a distinction between a manifest that's "new" because it's currently empty and a manifest that once had versions but is now empty? Wondering about the naming of "new"

src/server/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /// Starts the server, consuming it.
+    ///
+    /// This function will run until it either returns an error (which should always be fatal) or
+    /// you stop polling the future
+    #[instrument(level = "info", skip_all)]
+    pub async fn serve(mut self) -> anyhow::Result<()> {
+        while let Some(msg) = self.subscriber.next().await {
+            if !msg.subject.starts_with(&self.prefix) {
```
connorsmith256, line 70: Is this even possible in NATS, given that the subscriber has a topic prefix?

src/server/parser.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+/// Parse the bytes as yaml or json (in that order)
+fn parse_yaml_or_json(data: Vec<u8>) -> anyhow::Result<Manifest> {
+    serde_yaml::from_slice(&data).or_else(|e| {
+        serde_json::from_slice(&data).map_err(|err| {
+            // Combine both errors in case one was a legit parsing failure due to invalid data
+            anyhow::anyhow!("JSON parsing failed: {err:?}")
+                .context(format!("YAML parsing failed: {e:?}"))
```
connorsmith256, line 46: This feels a bit weird, given that we checked the content type header first. Why try to parse as JSON even if content-type was `application/yaml`, and vice versa?

src/server/handlers.rs
```diff
@@ -0,0 +1,347 @@
...
+            Ok(d) => d.unwrap_or_default(),
+            Err(e) => {
+                error!(error = %e, "Unable to fetch data from store");
+                self.send_error(msg.reply, "Internal storage error".to_string())
+                    .await;
+                return;
+            }
+        };
+
+        let current_version = manifest.version().to_owned();
```
connorsmith256, line 53: Naming: `current_version` made me think we'd be reading from `current_manifests`, not the incoming model

src/model/mod.rs
thomastaylor312: According to the OAM spec, it is perfectly valid to _not_ have a version. However, in wadm you could only have one version with `""`

bin/main.rs
thomastaylor312: We use our own prefix for things in Cosmonic all the time, so I figured it would be good to build it in. As for the bucket name, that could happen all the time for production as the guidance will be to preprovision your own buckets

src/server/mod.rs
thomastaylor312: Theoretically no, but I figured logging it would be a good idea, especially if someone did a custom prefix of some kind

src/server/parser.rs
thomastaylor312: This function of the fallback _only_ gets called if no content type header was set. The logic block above does the checking of the header and parsing

bin/main.rs
thomastaylor312: This is what is in code (which is correct)

```rust
let store = nats::ensure_kv_bucket(&context, args.state_bucket, 1).await?;

let state_storage = NatsKvStore::new(store);

let store = nats::ensure_kv_bucket(&context, args.manifest_bucket, 1).await?;

let manifest_storage = NatsKvStore::new(store);
```

src/model/mod.rs
```diff
@@ -12,9 +14,15 @@ pub const OAM_VERSION: &str = "core.oam.dev/v1beta1";
 // NOTE(thomastaylor312): If we ever end up supporting more than one kind, we should use an enum for
 // this
 pub const APPLICATION_KIND: &str = "Application";
+/// The version key, as predefined by the [OAM
+/// spec](https://github.com/oam-dev/spec/blob/master/metadata.md#annotations-format)
```
brooksmtownsend, line 18: just double checking, the line break here preserves the link?

src/model/mod.rs
brooksmtownsend: Did we want to opinionated pick something like `v0.0.1` like we did in wasmCloud?

src/server/handlers.rs
```diff
@@ -0,0 +1,388 @@
...
+                }
+            }
+            None => GetModelResponse {
+                manifest: Some(manifests.get_current().to_owned()),
+                result: GetResult::Success,
+                message: format!("Fetched model {name}"),
+            },
+        };
+        // SAFETY: We _just_ deserialized this from the store above, so we should be just fine
+        self.send_reply(msg.reply, serde_json::to_vec(&reply).unwrap())
```
brooksmtownsend, line 171: In general, is there a real downside to `unwrap_or_default` in all these safety cases?

src/server/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+    ) -> anyhow::Result<Server<S>> {
+        // Trim off any spaces or trailing/preceding dots
+        let prefix = topic_prefix
+            .unwrap_or(DEFAULT_WADM_TOPIC_PREFIX)
+            .trim()
+            .trim_matches('.')
+            .to_owned();
+        if prefix.is_empty() {
+            anyhow::bail!("Given prefix was empty")
+        }
```
brooksmtownsend, line 47: double checking, in the default case you expect the previs to be `wadm.api`, with all whitespace and leading/trailing periods removed?

brooksmtownsend (approved):
Only non-blocking comments/questions/quips

src/model/mod.rs
thomastaylor312: Yeah, it does as far as I know. This is just my wrapper thing

src/model/mod.rs
thomastaylor312: I'd rather be spec compliant here than pick something by default (especially because we don't enforce semver on it)

src/server/handlers.rs
thomastaylor312: Actually, there probably isn't and it would be more graceful. Let me go update those

src/server/mod.rs
thomastaylor312: Yes, that is correct

PR #73, opened 2023-04-04
feat(*): Handle deploy operations

thomastaylor312:
Please note that this doesn't try to do any sort of notification event when something is deployed. That will have to be handled later

### DIFF
--- a/src/server/handlers.rs
+++ b/src/server/handlers.rs
@@ -2,12 +2,16 @@ use async_nats::{Client, Message};
 use serde_json::json;
 use tracing::{debug, error, instrument, trace};
 
-use crate::{model::internal::StoredManifest, storage::Store};
+use crate::{
+    model::{internal::StoredManifest, LATEST_VERSION},
+    storage::Store,
+};
 
 use super::{
-    parser::parse_manifest, DeleteModelRequest, DeleteModelResponse, DeleteResult, GetModelRequest,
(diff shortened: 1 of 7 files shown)


src/server/handlers.rs
```diff
@@ -41,6 +45,15 @@ impl<S: Store + Send + Sync> Handler<S> {
             "Manifest is valid. Fetching current manifests from store"
         );
 
+        if manifest.version() == LATEST_VERSION {
+            self.send_error(
+                msg.reply,
+                format!("A manifest with a version {LATEST_VERSION} is not allowed in wadm"),
```
brooksmtownsend, line 51: Make sure to document this bad boy 😉

brooksmtownsend (approved):
Just like the breaking change to destructive undeploy, I like the idea of making `put` an operation that deploys by default, unless run with a `--put-only` or whatever command. Just feels nice to be able to do an essential `kubectl apply` with a manifest

src/server/handlers.rs
brooksmtownsend: No changes needed here. Just something we'll want to note when mentioning OAM versions

thomastaylor312:
@brooksmtownsend I actually thought about adding something about "put only" but then we have a separate request envelope we have to add. I'll add an issue as a follow up

PR #75, opened 2023-04-05
(feat) Implement the ActorSpreadScaler

### LINKED ISSUE #51: Fully implement the spreadscaler
As part of  we are implementing the basics of a spreadscaler. This work is to make sure everything is logged, instrumented, feature complete, and tested. It should implement all of the same functionality currently supported in wadm 0.3

brooksmtownsend:
Fixes #51 

This PR adds the `ActorSpreadScaler`, an implementation of the `Scaler` trait which spreads replicas of actors across a list of spread requirements. It does not attempt to evenly spread on hosts that match requirements, as that's more complicated and better suited for another Scaler implementation.

Currently marking as ready for review, but two (or three?) outstanding things I need to address
- [x] Host state is not modified in the store when an actor is stopped or started. Updated PR to ensure host state is updated when actors and providers start/stop
- [x] There may be an edge case where multiple spread requirements match on the same host, resulting in conflicting commands. Annotations will be required to solve this, may have to add support in this PR.
    - After rebasing and adding in #76, this functionality was added _and_ required a ton of work to refactor our current state to take into account instance IDs and annotations.
- [x] (future PR) This scaler ignores linkdef and provider events. If desired to show a similar implementation for those scalers, they will be in different files as different `Scaler` implementations

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Added multiple unit tests to ensure proper functionality of the spreadscaler

### Manual Verification
I ran the tests but did not run for reals since the mechanism for publishing commands isn't hooked up quite yet

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -18,7 +18,7 @@ macro_rules! from_impl {
 }
 
 /// All possible compensatory commands for a lattice
-#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 // In order to keep things clean, we are using untagged enum serialization. So it will just try to
 // match on data without tagging (see https://serde.rs/enum-representations.html for more info on
 // what the other options look like). These types are purely internal to wadm, but for greater
@@ -35,7 +35,7 @@ pub enum Command {
 }
 
 /// Struct for the StartActor command
-#[derive(Clone, Debug, Eq, Serialize, Deserialize, Default)]
+#[derive(Clone, Debug, Serialize, Deserialize, Default, Eq)]
 pub struct StartActor {
     /// The OCI or bindle reference to start
     pub reference: String,
@@ -45,25 +45,24 @@ pub struct StartActor {
     pub count: usize,
     /// The name of the model/manifest that generated this command
     pub model_name: String,
+    /// Additional annotations to attach on this command
+    pub annotations: HashMap<String, String>,
 }
 
 from_impl!(StartActor);
 
 impl PartialEq for StartActor {
-    fn eq(&self, other: &StartActor) -> bool {
-        self.reference == other.reference && self.host_id == other.host_id
-    }
-}
-
-impl Hash for StartActor {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        self.reference.hash(state);
-        self.host_id.hash(state);
+    fn eq(&self, other: &Self) -> bool {
+        self.reference == other.reference
+            && self.host_id == other.host_id
+            && self.count == other.count
+            && self.model_name == other.model_name
+            && self.annotations == other.annotations
     }
 }
 
 /// Struct for the StopActor command
-#[derive(Clone, Debug, Eq, Serialize, Deserialize, Default)]
+#[derive(Clone, Debug, Serialize, Deserialize, Default, Eq)]
 pub struct StopActor {
     /// The ID of the actor to stop
     pub actor_id: String,
@@ -73,20 +72,19 @@ pub struct StopActor {
     pub count: usize,
     /// The name of the model/manifest that generated this command
     pub model_name: String,
+    /// Additional annotations to attach on this command
+    pub annotations: HashMap<String, String>,
 }
 
 from_impl!(StopActor);
 
 impl PartialEq for StopActor {
-    fn eq(&self, other: &StopActor) -> bool {
-        self.actor_id == other.actor_id && self.host_id == other.host_id
-    }
-}
-
-impl Hash for StopActor {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        self.actor_id.hash(state);
-        self.host_id.hash(state);
+    fn eq(&self, other: &Self) -> bool {
+        self.actor_id == other.actor_id
+            && self.host_id == other.host_id
+            && self.count == other.count
+            && self.model_name == other.model_name
+            && self.annotations == other.annotations
     }
 }
 
--- a/src/scaler/spreadscaler.rs
+++ b/src/scaler/spreadscaler.rs
@@ -1,9 +1,8 @@
+use std::{cmp::Ordering, collections::HashMap};
+
 use anyhow::Result;
 use async_trait::async_trait;
-use std::{
-    cmp::Ordering,
-    collections::{HashMap, HashSet},
-};
+use tokio::sync::OnceCell;
 use tracing::log::warn;
 
 use crate::{
@@ -11,24 +10,36 @@ use crate::{
     events::{Event, HostStarted, HostStopped},
     model::{Spread, SpreadScalerProperty, DEFAULT_SPREAD_WEIGHT},
     scaler::Scaler,
-    storage::{Actor, Host, ReadStore},
+    storage::{Actor, Host, ReadStore, WadmActorInstance},
 };
 
+// Annotation constants
+const SCALER_KEY: &str = "wasmcloud.dev/scaler";
+const SCALER_VALUE: &str = "spreadscaler";
+const SPREAD_KEY: &str = "wasmcloud/dev/spread_name";
+
 /// Config for an ActorSpreadScaler
-struct ActorSpreadConfig {
+pub struct ActorSpreadConfig {
+    /// OCI, Bindle, or File reference for an actor
     actor_reference: String,
+    /// Lattice ID that this SpreadScaler monitors
     lattice_id: String,
+    /// The name of the wadm model this SpreadScaler is under
+    model_name: String,
+    /// Configuration for this SpreadScaler
     spread_config: SpreadScalerProperty,
+    /// Actor ID, stored in a OnceCell to facilitate more efficient fetches
+    actor_id: OnceCell<String>,
 }
 
-/// The SimpleScaler ensures that a certain number of replicas are running
-/// for a certain public key.
+/// The ActorSpreadScaler ensures that a certain number of replicas are running,
+/// spread across a number of hosts according to a [SpreadScalerProperty](crate::model::SpreadScalerProperty)
 ///
-/// This is primarily to demonstrate the functionality and ergonomics of the
-/// [Scaler](crate::scaler::Scaler) trait and doesn't make any guarantees
-/// about spreading replicas evenly
-struct ActorSpreadScaler<S: ReadStore + Send + Sync> {
+/// If no [Spreads](crate::model::Spread) are specified, this Scaler simply maintains the number of replicas
+/// on an available host
+pub struct ActorSpreadScaler<S: ReadStore + Send + Sync> {
     pub config: ActorSpreadConfig,
+    spread_requirements: Vec<(Spread, usize)>,
     store: S,
 }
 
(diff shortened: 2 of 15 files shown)


src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,316 @@
...
+struct ActorSpreadScaler<S: ReadStore + Send + Sync> {
+    pub config: ActorSpreadConfig,
+    store: S,
+}
+
+#[async_trait]
+impl<S: ReadStore + Send + Sync> Scaler for ActorSpreadScaler<S> {
+    type Config = ActorSpreadConfig;
+
+    fn update_config(&mut self, config: Self::Config) -> Result<bool> {
```
thomastaylor312, line 38: Was just thinking about this: Update config should probably trigger a reconcile

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+/// Given a spread config, return a vector of tuples that represents the spread
+/// and the actual number of actors to start for a specific spread requirement
+fn compute_spread(spread_config: &SpreadScalerProperty) -> Result<Vec<(&Spread, usize)>> {
+    // ideal return would be a Vec<(num_actors, Spread)>?
+    let replicas = spread_config.replicas;
+    // TODO: what should the default weight be? 100 would clobber anyone who uses 1, maybe default is the
+    // same as the minimum weight specified?
+    let default_weight: usize = 100;
```
thomastaylor312, line 110: Nit: We have a default weight constant

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,316 @@
...
+                    weight: None,
+                },
+            ],
+        };
+
+        let multi_spread_even_no_weight = compute_spread(&multi_spread_even_no_weight)?;
+        assert_eq!(multi_spread_even_no_weight[0].1, 5);
+        assert_eq!(multi_spread_even_no_weight[1].1, 5);
+
+        // Ensure we compute if weights are partially specified
```
thomastaylor312, line 265: Nit: I think these comments need to move and/or be modified

src/scaler/spreadscaler.rs
brooksmtownsend: Hm, true. Maybe this returns a HashSet of commands too

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,984 @@
...
+        for cmd in cmds.iter() {
+            match cmd {
+                Command::StartActor(start) => {
+                    assert_eq!(start.host_id, host_id_three.to_string());
+                    assert_eq!(start.count, 3);
+                    assert_eq!(start.reference, blobby_ref);
+                }
+                Command::StopActor(stop) => {
+                    assert_eq!(stop.host_id, host_id_two.to_string());
+                    assert_eq!(stop.count, 15);
```
brooksmtownsend, line 975: This assertion currently fails since the `ActorStopped` event doesn't modify the host state, which is what the spreadscaler reads from. Considering either refactoring the scaler (not desired) or adding host state modifications to the event worker (desired)

brooksmtownsend: edit: added modifications to event worker in https://github.com/wasmCloud/wadm/pull/75/commits/01a6194d0f954f825cae31fb34a021ce1b2604eb

src/workers/event.rs
```diff
@@ -870,8 +953,31 @@ mod test {
         assert_eq!(providers.len(), 2, "Should only be 2 providers in state");
         assert_provider(&providers, &provider2, &[&host1_id, &host2_id]);
 
-        // TODO(thomastaylor312): This is just a reminder that if we add in host state updating on
-        // provider/actor events, we should test that here before we hit a host heartbeat
+        // Check that hosts got updated properly
+        let hosts = store.list::<Host>(lattice_id).await.unwrap();
+        assert_eq!(hosts.len(), 2, "Should only have 2 hosts");
```
brooksmtownsend, line 958: General note that I essentially moved a few assertions _before_ the host heartbeat to ensure that the new functionality worked as well as leaving some assertions after the host heartbeat to ensure we handle the heartbeat properly too.

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,1197 @@
...
+
+        // NOTE(brooksmtownsend) it's easier to assign one host per list of requirements than
+        // balance within those requirements. Users should be specific with their requirements
+        // as wadm is not responsible for ambiguity, a future scaler like a DaemonScaler could handle this
+
+        let commands = spread_requirements
+            .iter()
+            .filter_map(|(spread, count)| {
+                let eligible_hosts = eligible_hosts(&hosts, spread);
+                if let Some(first_host) = eligible_hosts.get(0) {
```
thomastaylor312, line 106: Theoretically `eligible_hosts` _should_ return a different ordering of hosts every time because it is iterating over a hashmap to generate them. However, I worry about this favoring 1 host over another too much. For example, if I had two hosts that matched the requirement `foo: bar` and had those requirements on several manifests, I could get the same host multiple times for different models even though 2 match.

Now that I am thinking about this more, it could actually be a problem because the host you select is non-deterministic. On one call to reconcile, I could get Host A, and on a second call, I could get host B. We need to come up with a way to deterministically choose from a pool of hosts or spread across a pool of hosts. We _might_ be able to hold out on this until wadm 0.5, but we should talk about it. At a bare minimum, we should probably sort the vec of eligible hosts

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,1197 @@
...
+            .collect::<Vec<Command>>();
+
+        Ok(commands)
+    }
+}
+
+impl<S: ReadStore + Send + Sync> ActorSpreadScaler<S> {
+    #[allow(unused)]
+    /// Construct a new ActorSpreadScaler with specified configuration values
+    fn new(
```
thomastaylor312, line 171: Can you just `pub` this and remove the `allow(unused)` above?

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,1197 @@
...
+            .unwrap_or_default())
+    }
+
+    /// Helper function to create a predictable annotations map for a spread
+    fn annotations(&self, spread_name: &str) -> HashMap<String, String> {
+        HashMap::from_iter([
+            (SCALER_KEY.to_string(), SCALER_VALUE.to_string()),
+            (SPREAD_KEY.to_string(), spread_name.to_string()),
+            (
+                crate::MANAGED_BY_ANNOTATION.to_owned(),
```
thomastaylor312, line 211: These should just be handled by the command worker. It is probably better to modify that logic to merge in these values if they aren't set. I also think the scaler/model name should be on the commands and then get set automatically by the command handler to ensure we have a way to look up which scaler to use

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,1197 @@
...
+                },
+            ],
+        };
+
+        let multi_spread_even_no_weight = compute_spread(&multi_spread_even_no_weight)?;
+        assert_eq!(multi_spread_even_no_weight[0].1, 5);
+        assert_eq!(multi_spread_even_no_weight[1].1, 5);
+
+        // Ensure we compute if weights are partially specified
+        // Ensure we compute if spread vec is empty?
```
thomastaylor312, line 414: Nit: Can remove the `?` here now

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,1197 @@
...
+                                annotations: echo_spreadscaler.annotations("CrossRegionReal"),
+                            })),
+                        ),
+                    ]),
+                    reference: blobby_ref.to_string(),
+                },
+            )
+            .await?;
+
+        // STATE SETUP BEGIN
```
thomastaylor312, line 725: Super nit: Copy paste from above that you don't need?

src/storage/reaper.rs
```diff
@@ -164,13 +164,15 @@ impl<S: Store + Clone + Send + Sync + 'static> Undertaker<S> {
             actors
                 .into_iter()
                 .filter_map(|(id, mut actor)| {
-                    let current_num_hosts = actor.count.len();
+                    let current_num_hosts = actor.instances.len();
                     // Only keep the instances where the host exists
-                    actor.count.retain(|host_id, _| hosts.contains_key(host_id));
+                    actor
```
thomastaylor312, line 169: This stuff reminded me that we probably need a `NodeReaped` event so we can trigger reconciliation operations

src/storage/state.rs
```diff
@@ -259,3 +287,37 @@ impl From<&HostHeartbeat> for Host {
         }
     }
 }
+
+// Custom Hash impl's due to nested HashMaps and HashSets
+impl Hash for WadmActorInstance {
```
thomastaylor312, line 292: Nit: Can you move these impls up next to the other impls for this struct?

src/storage/state.rs
```diff
@@ -259,3 +287,37 @@ impl From<&HostHeartbeat> for Host {
...
+}
+
+/// NOTE(brooksmtownsend): This is compared only by the instance ID
+/// as it's a GUID and lookups don't require annotations
+impl PartialEq for WadmActorInstance {
+    fn eq(&self, other: &Self) -> bool {
+        self.instance_id == other.instance_id
+    }
+}
+impl Eq for WadmActorInstance {}
```
thomastaylor312, line 305: Nit: If you wanna be lazy, you can just add `Eq` to the `derive`

src/storage/state.rs
```diff
@@ -259,3 +287,37 @@ impl From<&HostHeartbeat> for Host {
...
+impl Hash for Actor {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.id.hash(state);
+        self.name.hash(state);
+        self.capabilities.hash(state);
+        self.issuer.hash(state);
+        self.call_alias.hash(state);
+        self.reference.hash(state);
+
+        for (host_id, instances) in &self.instances {
```
thomastaylor312, line 316: Same thing here with non deterministic hashing. However, does `Actor` need to be `Hash`? We aren't storing it in a HashSet anywhere, are we?

src/scaler/spreadscaler.rs
thomastaylor312: Update after reading further: Looks like you actually already did this, so these can be removed. My comment about having the command handler automatically label the name seems like a better idea for us so we can look things up

src/workers/event.rs
```diff
@@ -114,15 +125,42 @@ where
         {
             trace!(actor = ?current, "Found existing actor data");
             // Merge in current counts
-            actor_data.count = current.count;
+            actor_data.instances = current.instances;
+        }
+        // Update actor count in the host
+        if let Some(mut host) = self.store.get::<Host>(lattice_id, &actor.host_id).await? {
+            trace!(host = ?host, "Found existing host data");
+            match host.actors.get(&actor.public_key) {
```
thomastaylor312, line 133: Nit: Consider using the [entry api](https://doc.rust-lang.org/1.68.2/std/collections/hash_map/struct.HashMap.html#method.entry) here instead

src/workers/event.rs
```diff
@@ -166,6 +210,27 @@ where
...
+                Some(existing_count) if *existing_count <= 1 => {
+                    host.actors.remove(&actor.public_key);
+                }
+                Some(existing_count) => {
+                    host.actors
+                        .insert(actor.public_key.to_owned(), *existing_count - 1);
+                }
+                // you cannot delete what doesn't exist
```
thomastaylor312, line 225: tenor-31310621

thomastaylor312 (changes requested):
Excellent testing and work here. There are some potential issues with the scaling algorithm we should discuss

src/commands/mod.rs
```diff
@@ -45,25 +45,38 @@ pub struct StartActor {
...
-}
-
 impl Hash for StartActor {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.reference.hash(state);
         self.host_id.hash(state);
+        self.count.hash(state);
+        self.model_name.hash(state);
+
+        for annotation in &self.annotations {
```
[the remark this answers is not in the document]
brooksmtownsend: Going to noodle on this 🤔 We may be able to get away with not hashing annotations if we aren't actually hashing this command anymore (after using Vec<Command>'s)

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,1197 @@
...
+            }
+            // No other event impacts the job of this scaler so we can ignore it
+            _ => Ok(Vec::new()),
+        }
+    }
+
+    async fn reconcile(&self) -> Result<Vec<Command>> {
+        let spread_requirements = compute_spread(&self.config.spread_config)?;
+        let hosts = self.store.list::<Host>(&self.config.lattice_id).await?;
+        let actor_id = self.actor_id().await?;
```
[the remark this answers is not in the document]
brooksmtownsend: That is true, we could fetch that on initialization and/or when we update the config

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,1197 @@
...
+                    self.reconcile().await
+                }
+            }
+            // No other event impacts the job of this scaler so we can ignore it
+            _ => Ok(Vec::new()),
+        }
+    }
+
+    async fn reconcile(&self) -> Result<Vec<Command>> {
+        let spread_requirements = compute_spread(&self.config.spread_config)?;
```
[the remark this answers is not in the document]
brooksmtownsend: Will look to add this along with the actor ID

src/scaler/spreadscaler.rs
brooksmtownsend: That is true, I thought one of my test cases covered this but the vec of spreads is ordered while the map of hosts isn't 🤔 

I liked the idea of it being random just to not favor one host too much, and I actually think non-determinism is okay here. Essentially here's what would possibly happen:

1. Define spread that asks for 12 replicas of an actor, 10 are running across the lattice on eligible hosts
2. We ask to start 2 more instances on host A
3. we start 2 more instances on host A
4. Next call to reconcile finds 12 actors on eligible hosts, so no commands needed
5. Sometime later, user deletes 5 instances from some host
6. We find 7 running actors on eligible hosts
7. We ask to start 5 more instances on host A, or maybe B, or maybe C
8. we start 5 more
9. next call to reconcile finds nothing to do

e.g. I think it's not a problem here and the randomness is kinda okay. Do you think it would cause problems?

src/scaler/spreadscaler.rs
brooksmtownsend: This doesn't seem to resolve the warning 🤔

src/storage/state.rs
brooksmtownsend: Huh, looks like we aren't. Will remove

src/workers/event.rs
brooksmtownsend: FINE. I have to clone but I'll do it

src/scaler/spreadscaler.rs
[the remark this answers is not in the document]
brooksmtownsend: Hm, could do this but if the actor hasn't started before we won't be able to fetch the ID 🤔 Then I'd need a mutable self during reconcile which feels wrong

src/commands/mod.rs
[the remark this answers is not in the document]
brooksmtownsend: I think, simply, at this point we aren't storing the commands in a HashMap/HashSet anywhere, so we can just remove the derived and implemented traits since we're working with vecs. Will want to revisit if that's the case, but I don't see a reason to change our data structure to a BTreeMap or an IndexMap just to support a use case we don't need right now

src/commands/mod.rs
```diff
@@ -73,22 +86,35 @@ pub struct StopActor {
...
-}
-
 impl Hash for StopActor {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.actor_id.hash(state);
         self.host_id.hash(state);
+        self.count.hash(state);
+        self.model_name.hash(state);
+
+        for annotation in &self.annotations {
```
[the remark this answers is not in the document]
brooksmtownsend: See above comment for proposed suggestion

src/scaler/spreadscaler.rs
```diff
@@ -0,0 +1,1197 @@
...
+                    spread.requirements.iter().all(|(key, value)| {
+                        labels.get(key).map(|val| val == value).unwrap_or(false)
+                    })
+                }) {
+                    Ok(Vec::new())
+                } else {
+                    self.reconcile().await
+                }
+            }
+            // No other event impacts the job of this scaler so we can ignore it
```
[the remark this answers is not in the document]
thomastaylor312: Leaving this as unresolved so I can see it when I come back later

src/scaler/spreadscaler.rs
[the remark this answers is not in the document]
thomastaylor312: What if we did a lazy initialization? So we'd still use the `actor_id()` function, but it could wrap a `OnceCell` so that we only ever fetch it once

src/scaler/spreadscaler.rs
thomastaylor312: What if the count for this spread is across several nodes? Right now this only gets the count on the first node. Example logic

eligible hosts: [HostA, HostB]
replicas: 10

1. First reconcile chooses HostA to put 10 actors on it
2. 3 actors are stopped on HostA
3. Second reconcile schedules on HostB, sees a current count of actors of 0, schedules 10 actors on HostB
4. Now 20 replicas are running

Am I reading the logic right?

src/scaler/spreadscaler.rs
thomastaylor312: Is `ActorSpreadScaler` publicly exported in the `scaler` module?

src/scaler/spreadscaler.rs
thomastaylor312: I think the model name should be an externally set annotation since we'll use it to key on. The spread name is internal to the scaler implementation

src/storage/reaper.rs
thomastaylor312: Created #80 to follow up on this

src/scaler/spreadscaler.rs
thomastaylor312: After talking to @brooksmtownsend, I was reading the logic wrong. This block just selects a host and _then_ checks all running instances of the actor

src/scaler/spreadscaler.rs
thomastaylor312: After talking to @brooksmtownsend the only thing we need to do here is remove the managed by annotation since the main wadm code takes care of it

thomastaylor312 (approved):

PR #76, opened 2023-04-05
feat(workers): Adds a new command worker

### LINKED ISSUE #50: Actually handle commands in the work loop
As part of Stage 1, we are implementing a dummy work loop to handle commands. This task involves taking a list of commands and turning them into actual compensatory actions against the lattice control topics

thomastaylor312:
This adds a worker that actually translates commands into lattice ctl operations. It also starts the work of creating a control client per lattice as needed

Closes #50

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,3 +20,9 @@ pub const DEFAULT_EXPIRY_TIME: Duration = Duration::from_secs(70);
 pub const DEFAULT_EVENTS_TOPIC: &str = "wasmbus.evt.*";
 /// Default topic to listen to for all commands
 pub const DEFAULT_COMMANDS_TOPIC: &str = "wadm.cmd.*";
+/// Managed by annotation used for labeling things properly in wadm
+pub const MANAGED_BY_ANNOTATION: &str = "wasmcloud.com/managed-by";
+/// Identifier for managed by annotation. This is the value [`MANAGED_BY_ANNOTATION`] is set to
+pub const MANAGED_BY_IDENTIFIER: &str = "wadm";
+/// The default link name. In the future, this will likely be pulled in from another crate
+pub const DEFAULT_LINK_NAME: &str = "default";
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -19,3 +19,9 @@ pub const DEFAULT_EXPIRY_TIME: Duration = Duration::from_secs(70);
 pub const DEFAULT_EVENTS_TOPIC: &str = "wasmbus.evt.*";
 /// Default topic to listen to for all commands
 pub const DEFAULT_COMMANDS_TOPIC: &str = "wadm.cmd.*";
+/// Managed by annotation used for labeling things properly in wadm
+pub const MANAGED_BY_ANNOTATION: &str = "wasmcloud.com/managed-by";
```
autodidaddict, line 23: The old (current) wadm didn't need this. It just used the "spec" annotation to indicate which oam model owned a particular instance of a thing. If the spec didn't exist then it was considered unmanaged.

Is there a different consideration here that needs a managed by as well as the claimant spec?

autodidaddict (commented):
This looks good but I'd like to see integration tests for when unmanaged instances of an actor are on the same host as managed instances to verify that the unmanaged ones remain unaffected and reconciliation done by wadm doesn't affect unmanaged actors or providers.

autodidaddict (approved):

src/lib.rs
thomastaylor312: Oh I actually think that is a better idea. I'll go update

thomastaylor312:
> This looks good but I'd like to see integration tests for when unmanaged instances of an actor are on the same host as managed instances to verify that the unmanaged ones remain unaffected and reconciliation done by wadm doesn't affect unmanaged actors or providers.

I did test this 🙂. It is the last test done in the integration test

src/lib.rs
autodidaddict: Note that there is a specific annotation already detected by the lattice observer used by the current wadm.

PR #77, opened 2023-04-07
feat(*): Automatically observe lattices

thomastaylor312:
This commit makes it so that wadm automatically watches for new lattices to appear and then manages them. As part of this, it made sense to actually make sure this is mutlitenant ready as well. So this implements the functionality to properly observe for multitenant (read: cross account importing) architectures.

Not gonna lie that some of the stuff to start watching pre-observed lattices is a bit messy, but I did test it out and it is real slick

### DIFF
--- a/bin/observer.rs
+++ b/bin/observer.rs
@@ -1,0 +1,113 @@
+//! Types for observing a nats cluster for new lattices
+
+use async_nats::Subscriber;
+use futures::{stream::SelectAll, StreamExt, TryFutureExt};
+use tracing::{debug, error, instrument, trace, warn};
+
+use wadm::{
+    consumers::{manager::ConsumerManager, CommandConsumer, EventConsumer},
+    events::{EventType, HostHeartbeat, HostStarted},
+    nats_utils::LatticeIdParser,
+    storage::{nats_kv::NatsKvStore, reaper::Reaper, Store},
+    workers::{CommandWorker, EventWorker},
+    DEFAULT_COMMANDS_TOPIC,
+};
+
+use super::connections::ControlClientConstructor;
+
+pub(crate) struct Observer<S> {
+    pub(crate) client_builder: ControlClientConstructor,
+    pub(crate) parser: LatticeIdParser,
+    pub(crate) command_manager: ConsumerManager<CommandConsumer>,
+    pub(crate) event_manager: ConsumerManager<EventConsumer>,
+    pub(crate) client: async_nats::Client,
+    pub(crate) store: S,
+    pub(crate) reaper: Reaper<NatsKvStore>,
+}
+
+impl<S: Store + Send + Sync + Clone + 'static> Observer<S> {
+    /// Watches the given topic (with wildcards) for wasmbus events. If it finds a lattice that it
+    /// isn't managing, it will start managing it immediately
+    ///
+    /// If this errors, it should be considered fatal
+    #[instrument(level = "info", skip(self))]
+    pub(crate) async fn observe(mut self, subscribe_topics: Vec<String>) -> anyhow::Result<()> {
+        let mut sub = get_subscriber(&self.client, subscribe_topics.clone()).await?;
+        loop {
+            match sub.next().await {
+                Some(msg) => {
+                    if !is_event_we_care_about(&msg.payload) {
+                        continue;
+                    }
+                    let lattice_id = match self.parser.parse(&msg.subject) {
+                        Some(id) => id,
+                        None => {
+                            trace!(subject = %msg.subject, "Found non-matching lattice subject");
+                            continue;
+                        }
+                    };
+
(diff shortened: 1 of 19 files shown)


src/consumers/manager.rs
```diff
@@ -88,22 +99,85 @@ impl<C> ConsumerManager<C> {
...
+        let mut manager = ConsumerManager {
             handles: Arc::new(RwLock::new(HashMap::default())),
             permits: permit_pool,
             stream,
             phantom: PhantomData,
-        }
+        };
+
+        let handles: HashMap<String, JoinHandle<WorkResult<()>>> = manager.stream.consumers().filter_map(|res| async {
+            let info = match res {
```
connorsmith256, line 128: I think you could use the fancy `if let Ok... else` syntax here yeah?

src/consumers/manager.rs
```diff
@@ -88,22 +99,85 @@ impl<C> ConsumerManager<C> {
...
+            // getting the last thing in case of other underscores
+            let lattice_id = match info.name.split('_').last() {
+                Some(id) => id,
+                None => return None,
+            };
+            // NOTE(thomastaylor312): It might be nicer for logs if we add an extra param for a
+            // friendly consumer manager name
+            trace!(%lattice_id, "Adding consumer for lattice");
+
+            let worker = match worker_generator.create(lattice_id).await {
```
connorsmith256, line 147: Use fancy?

src/consumers/manager.rs
```diff
@@ -88,22 +99,85 @@ impl<C> ConsumerManager<C> {
...
+                Ok(w) => w,
+                Err(e) => {
+                    error!(error = %e, %lattice_id, "Unable to add consumer for lattice. Error when generating worker");
+                    return None;
+                }
+            };
+
+            // NOTE: If this becomes a bottle neck at scale, we could rewrite to take a single write lock and then
```
connorsmith256, line 155: And then.... And then what?? 😉

src/consumers/mod.rs
```diff
@@ -87,7 +87,7 @@ impl<T> ScopedMessage<T> {
     /// This method doesn't have an error because if you nack, it means the message needs to be
     /// rehandled, so the ack wait will eventually expire
     pub async fn nack(&mut self) {
-        if let Err(e) = self.custom_ack(AckKind::Nak).await {
+        if let Err(e) = self.custom_ack(AckKind::Nak(None)).await {
```
connorsmith256, line 90: What's this? I don't see anything contained in the enum value upstream

bin/observer.rs
```diff
@@ -0,0 +1,113 @@
...
+                    sub = get_subscriber(&self.client, subscribe_topics.clone()).await?;
+                }
+            }
+        }
+    }
+}
+
+// This is a stupid hacky function to check that this is a host started or host heartbeat event
+// without actually parsing
+fn is_event_we_care_about(data: &[u8]) -> bool {
```
connorsmith256, line 91: Did you just not want to depend on `cloud_event`? I think we've done `serde_json::from_slice::<CloudEvent>(data).and_then(|evt| evt.ty())` in other places

bin/observer.rs
```diff
@@ -0,0 +1,113 @@
...
+    #[instrument(level = "info", skip(self))]
+    pub(crate) async fn observe(mut self, subscribe_topics: Vec<String>) -> anyhow::Result<()> {
+        let mut sub = get_subscriber(&self.client, subscribe_topics.clone()).await?;
+        loop {
+            match sub.next().await {
+                Some(msg) => {
+                    if !is_event_we_care_about(&msg.payload) {
+                        continue;
+                    }
+                    let lattice_id = match self.parser.parse(&msg.subject) {
```
connorsmith256, line 42: Fancaaay syntax?

connorsmith256 (approved):
LGTM, just a couple questions

src/consumers/manager.rs
thomastaylor312: Yeah, it was mostly a style choice to use match. I think I started with `if let` and didn't like it as much

src/consumers/mod.rs
thomastaylor312: This is from the latest 0.29 nats. It now takes a duration to wait before sending NAK

bin/observer.rs
thomastaylor312: Was trying to avoid the penalty of parsing and allocating just to check type

src/consumers/manager.rs
thomastaylor312: Oh wait, can't capture the error that way

src/consumers/manager.rs
thomastaylor312: And then...I changed my mind and did what I was writing anyway 🤣

PR #79, opened 2023-04-11
feat(*): Notifies on manifest deployment

### LINKED ISSUE #49: Create work loop that handles incoming manifests
We need to have a loop that triggers on watching the KV store for manifest changes. This should be an easy thing to add to the struct that stores manifests. This should take any changes and automatically trigger an update from the appropriate Scaler

thomastaylor312:
Please note that this is a bit big because I had to have a stream that combined internal wadm events with wasmbus ones. However, I did realize that we only care about streams for their configuration and delivery guarantees, and not for their storage capabilities. Because of this, we don't actually have to have the users configure the streams beforehand. So while this adds one more stream to manage combining the two topics (until nats 2.10 comes out and we move to it), the user no longer has to configure any of them.

Closes #49

### DIFF
--- a/bin/main.rs
+++ b/bin/main.rs
@@ -10,11 +10,13 @@ use wadm::{
         manager::{ConsumerManager, WorkerCreator},
         *,
     },
+    mirror::Mirror,
     nats_utils::LatticeIdParser,
-    server::{Server, DEFAULT_WADM_TOPIC_PREFIX},
+    server::{Server, StreamNotifier, DEFAULT_WADM_TOPIC_PREFIX},
     storage::{nats_kv::NatsKvStore, reaper::Reaper},
     workers::{CommandWorker, EventWorker},
     DEFAULT_COMMANDS_TOPIC, DEFAULT_EVENTS_TOPIC, DEFAULT_MULTITENANT_EVENTS_TOPIC,
+    DEFAULT_WADM_EVENTS_TOPIC,
 };
 
 mod connections;
@@ -23,6 +25,10 @@ mod nats;
 mod observer;
 
 use connections::{ControlClientConfig, ControlClientConstructor};
+
+const EVENT_STREAM_NAME: &str = "wadm_events";
+const COMMAND_STREAM_NAME: &str = "wadm_commands";
+const MIRROR_STREAM_NAME: &str = "wadm_mirror";
 
 #[derive(Parser, Debug)]
 #[command(name = clap::crate_name!(), version = clap::crate_version!(), about = "wasmCloud Application Deployment Manager", long_about = None)]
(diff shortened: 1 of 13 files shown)


src/mirror/mod.rs
```diff
@@ -0,0 +1,143 @@
+//! This is a temporary workaround to let us combine multiple topics into a single stream so
+//! consumers work properly. This will be removed once NATS 2.10 is out and we have upgraded to
+//! using it in wasmcloud projects
```
brooksmtownsend, line 3: Would suggest issue for this

src/mirror/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+            .stream
+            .get_or_create_consumer(
+                &consumer_name,
+                PullConfig {
+                    durable_name: Some(consumer_name.clone()),
+                    name: Some(consumer_name.clone()),
+                    description: Some(format!("Durable wadm mirror consumer for {lattice_id}")),
+                    ack_policy: async_nats::jetstream::consumer::AckPolicy::Explicit,
+                    // These should be real quick to ack
+                    ack_wait: std::time::Duration::from_secs(1),
```
brooksmtownsend, line 60: Since this is a mirror, it's always a local NATS message right? I'd only ask to make this configurable if this message will traverse over a real network boundary

src/server/handlers.rs
```diff
@@ -369,6 +376,31 @@ impl<S: Store + Send + Sync> Handler<S> {
...
+        // an odd state. So I'd rather err on the side of caution and send a notification that gets
+        // ignored
+        if reply_data.undeploy || matches!(reply_data.result, DeleteResult::Noop) {
+            trace!("Sending undeploy notification");
+            if let Err(e) = self.notifier.undeployed(lattice_id, name).await {
+                error!(error = ?e, "Error when attempting to send undeploy notification during delete");
+                self.send_reply(
+                    msg.reply,
+                    // NOTE: We are constructing all data here, so this shouldn't fail, but just in
+                    // case we unwrap to nothing
```
brooksmtownsend, line 391: I like this safety pattern

src/server/handlers.rs
```diff
@@ -438,6 +470,12 @@ impl<S: Store + Send + Sync> Handler<S> {
             .await;
             return;
         }
+        // SAFETY: We can unwrap here because we know we _just_ successfully deployed the manifest so they should all exist
+        let manifest = manifests
+            .get_version(manifests.deployed_version().unwrap())
+            .unwrap()
+            .to_owned();
```
brooksmtownsend, line 477: Should is a strong word sir

src/server/notifier.rs
```diff
@@ -0,0 +1,99 @@
...
+pub struct StreamNotifier {
+    prefix: String,
+    context: Context,
+}
+
+impl StreamNotifier {
+    /// Creates a new stream notifier with the given prefix. This prefix should be something like
+    /// `wadm.evt` that is used to form the full topic to send to
+    pub fn new(prefix: &str, context: Context) -> StreamNotifier {
+        let trimmer: &[_] = &['.', '>'];
```
brooksmtownsend, line 40: 🤔 Shouldn't this include `*` too

bin/main.rs
```diff
@@ -256,17 +251,27 @@ async fn main() -> anyhow::Result<()> {
         [],
     );
 
+    let trimmer: &[_] = &['.', '*'];
```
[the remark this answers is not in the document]
thomastaylor312: Right now no. In the future I don't think so either as these topics should always end with the lattice id

src/events/types.rs
```diff
@@ -95,6 +97,10 @@ pub enum Event {
     HostHeartbeat(HostHeartbeat),
     LinkdefSet(LinkdefSet),
     LinkdefDeleted(LinkdefDeleted),
+    // NOTE(thomastaylor312): We may change where and how these get published, but it makes sense
+    // for now
+    ManifestPublished(ManifestPublished),
+    ManifestUnpublished(ManifestUnpublished),
```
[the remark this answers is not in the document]
thomastaylor312: I'll add to the comment above

src/mirror/mod.rs
thomastaylor312: It will likely cross a network boundary. I'll bump this to 2s, but because this is temporary I don't really see a need to make it configurable

brooksmtownsend (approved):

PR #84, opened 2023-04-15
feat(*) Link "spreadscaler" implementation

brooksmtownsend:
## Feature or Problem
This PR implements a Link spreadscaler, which simply waits until it can find the proper actor and provider IDs before publishing a link. Currently this does not check to see if the link exists already, which is why this is marked as draft and there's a TODO comment

## Related Issues
#51 

## Release Information
0.4.0

## Consumer Impact
Links!

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Looking to add basic unit tests to validate functionality

PR #85, opened 2023-04-15
feat(*): Integrates in the scalers into the event workers

### LINKED ISSUE #49: Create work loop that handles incoming manifests

thomastaylor312:
This is a giant PR because a lot of debugging happened. Most of this stuff is Real Code we will keep around. However, there is one big hacky thing to point out. In order to handle Actor Start/Stop events, we had to introduce the idea of a backoff, similar to what is done in wadm 0.3. This method definitely puts the "eventual" in eventual consistency, particularly with high replica counts, but it does work. This backoff is not needed for linkdef or provider scalers. The reason why this was needed was due to there being no way for identifying "groups" (for lack of a better term) of started and stopped events. We ended up with a stream of events with a non-deterministic end. The real solution here would be to update our actor started/stopped events to just emit a single event when a requested scale is done. So if a command comes in to start 10 actors, then we get all the same stuff we current do with a new `count` field and sans the instance-id field, which I don't think we use anywhere except the host and lattice observer (and I think the observer can do without it). This is also better for state storage as we don't have to read and write for _every_ single actor that starts or stops

Closes #49

### DIFF
--- a/src/scaler/mod.rs
+++ b/src/scaler/mod.rs
@@ -1,8 +1,10 @@
 use anyhow::Result;
 use async_trait::async_trait;
+use tokio::sync::mpsc::Sender;
 
-use crate::{commands::Command, events::Event};
+use crate::{commands::Command, events::Event, model::TraitProperty};
 
+pub mod manager;
 mod simplescaler;
 pub mod spreadscaler;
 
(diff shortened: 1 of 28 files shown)


src/scaler/mod.rs
```diff
@@ -18,15 +20,37 @@ pub mod spreadscaler;
...
+    /// be careful when calling backoff until you have sent the commands to their intended location
+    /// as otherwise the might not rerun until they are reconciled again.
+    ///
+    /// This function should return as quickly as possible, spawning a task that can switch the
+    /// state back from backoff. The passed in channel is meant to be used to notify the caller that
+    /// the backoff is complete with the name of the model. This name can then be used by the caller
+    /// to run a reconciliation pass to ensure everything is done. This is intended to workaround
+    /// the fact that there is no way to tell when we finished stopping/starting actors
+    ///
+    /// Please note that we'd like this to be temporary in the long term
```
brooksmtownsend, line 54: I think this will be the hardest thing for other folks to do _correctly_ if they end up creating scalers to extend wadm, and I definitely like the idea of us dropping this. Putting this here so I remember to put this in #86

src/workers/event.rs
```diff
@@ -700,69 +646,188 @@ where
...
+                .map(|_| None),
+            Event::LinkdefDeleted(_ld) => Ok(None),
+            Event::ProviderStarted(provider) => self
+                .handle_provider_started(&message.lattice_id, provider)
+                .await
+                .map(|_| {
+                    provider
+                        .annotations
+                        .get(APP_SPEC_ANNOTATION)
+                        .map(|s| s.as_str())
```
brooksmtownsend, line 781: note for myself: make sure provider scaler PR adds in custom annotations too

brooksmtownsend: This will go in the command section but I just remembered

src/workers/event.rs
```diff
@@ -700,69 +646,188 @@ where
...
+                .map(|_| None),
+            Event::HostStarted(host) => self
+                .handle_host_started(&message.lattice_id, host)
+                .await
+                .map(|_| None),
+            Event::HostStopped(host) => self
+                .handle_host_stopped(&message.lattice_id, host)
+                .await
+                .map(|_| None),
+            Event::LinkdefDeleted(_ld) => Ok(None),
```
brooksmtownsend, line 773: We will need to care about this with the linkdef scaler, but I can add it in #84

brooksmtownsend (approved):
Approved, but two things

1. The backoff function on the Scaler trait is stinky and we should definitely look to do host modifications where possible, and if it's still necessary manage backoff in the ScalerManager itself. It's not _too_ bad, but when implementing a scaler it's just enough to turn a nice little functional trait into having some spaghetti
2. I think wadm would do really well with a developer architecture guide, especially with a 1-cluster and a 3-cluster diagram, in terms of all the workers and traits that we use and how they communicate with each other. Not needed now of course, but maybe we can create an issue for it

PR #92, opened 2023-04-17
feat(*) simple devcontainer

brooksmtownsend:
This PR adds a simple devcontainer that installs `wash` and `wadm` so that people can run through the example without installing anything locally, if they prefer. Additionally it adds an "Open in Codespaces" badge, which is slick.

We install `wash` directly using `cargo` as we have some commands we need to change + release in wash, but wadm is downloaded from the github release directly

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,6 +6,9 @@
 			"version": "latest"
 		}
 	},
+  "containerEnv": {
+		"RUST_LOG": "INFO"
+	},
 	"postCreateCommand": "bash ./.devcontainer/install.sh",
 	"workspaceMount": "source=${localWorkspaceFolder},target=/wadm,type=bind,consistency=cached",
 	"workspaceFolder": "/wadm"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,13 +1,28 @@
 {
 	"name": "wadm",
-	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
+	"image": "mcr.microsoft.com/devcontainers/rust:latest",
 	"features": {
-		"ghcr.io/devcontainers/features/rust:1": {
-			"version": "latest"
-		}
+		"ghcr.io/devcontainers/features/common-utils:2": {}
 	},
-  "containerEnv": {
+	"containerEnv": {
 		"RUST_LOG": "INFO"
+	},
+	"customizations": {
+		"vscode": {
+			"settings": {
+				"files.watcherExclude": {
+					"**/target/**": true
+				},
+				"[rust]": {
+					"editor.formatOnSave": true
+				}
+			},
+			"extensions": [
+				"rust-lang.rust-analyzer",
+				"tamasfe.even-better-toml",
+				"serayuzgur.crates"
+			]
+		}
 	},
 	"postCreateCommand": "bash ./.devcontainer/install.sh",
 	"workspaceMount": "source=${localWorkspaceFolder},target=/wadm,type=bind,consistency=cached",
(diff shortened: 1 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
+{
+	"name": "wadm",
+	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
+	"features": {
+		"ghcr.io/devcontainers/features/rust:1": {
```
stephenandary, line 5: If this devcontainer is meant to work in, you may want to include a few more features like: 
- <https://github.com/devcontainers/features/tree/main/src/common-utils>

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
+{
+	"name": "wadm",
+	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
```
stephenandary, line 3: You can consider using rust image base instead of adding rust feature if just using some debian base: 
- <https://github.com/devcontainers/images/tree/main/src/rust>

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
...
+	"features": {
+		"ghcr.io/devcontainers/features/rust:1": {
+			"version": "latest"
+		}
+	},
+  "containerEnv": {
+		"RUST_LOG": "INFO"
+	},
+	"postCreateCommand": "bash ./.devcontainer/install.sh",
```
stephenandary, line 12: Something that I do for my dev containers is use another repo to build the image with CI, and then just reference the image here instead. I can help with that if you wanted, but if using small dev containers it shouldn't be such a significant difference. When installing a bunch with Cargo, Brew, or anything else that takes a long time it it makes a big difference. postCreateCommands will not be a part of image when you do that though, so good to note it is better to make this a local feature in those cases.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
...
+	"workspaceMount": "source=${localWorkspaceFolder},target=/wadm,type=bind,consistency=cached",
+	"workspaceFolder": "/wadm"
+}
```
stephenandary, line 15: If there are any extensions or settings you want to use, or to use a non-root user you can include as: 
	"customizations": {
		"vscode": {
			"settings": {
				"lldb.executable": "/usr/bin/lldb",
				// Tell VSCode not to watch target dir
				"files.watcherExclude": {
					"**/target/**": true
				},
				"rust-analyzer.checkOnSave.command": "clippy",
				"[rust]": {
					"editor.formatOnSave": true
				}
			},
			"extensions": [
				"vadimcn.vscode-lldb",
				"mutantdino.resourcemonitor",
				"rust-lang.rust-analyzer",
				"tamasfe.even-better-toml",
				"serayuzgur.crates"
			]
		}
	},
	// Use 'forwardPorts' to make a list of ports inside the container available locally.
	// "forwardPorts": [],
	// Use 'postCreateCommand' to run commands after the container is created.
	// "postCreateCommand": "",
	// Set `remoteUser` to `root` to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
	"remoteUser": "vscode"

.devcontainer/devcontainer.json
brooksmtownsend: I don't really like it being a part of the postCreate here, so I probably will just move it to a local feature. I think I did it this way because I got annoyed with the feature install being run as root and then the container couldn't find the proper binaries

brooksmtownsend: Eventually this will get a lot quicker though when we can `apt install wash` instead of building from source

thomastaylor312 (approved):

Issue #102, opened 2023-04-24
[FEATURE] Support `config_json` for provider start commands


brooksmtownsend:
Capability providers support an optional JSON configuration which we should be able to support via a `config` block in wadm. You can see the `configuration` variable in the interface: https://docs.rs/wasmcloud-interface-lattice-control/0.18.0/wasmcloud_interface_lattice_control/struct.StartProviderCommand.html. 

This is most useful for defining configuration that a provider will use for _every_ link, like a different default HTTPserver port with the wasmCloud httpserver provider. This `config` block would be taken and sent along to the start provider command. It would be nicer to specify nested config as YAML or JSON instead of just an opaque string, but the provider itself may or may not support a specific format

This would look something like this:
```
    - name: httpserver
      type: capability
      properties:
        image: wasmcloud.azurecr.io/httpserver:0.17.0
        contract: wasmcloud:httpserver
        config: {"address": "0.0.0.0:8085"}
      traits:
        - type: spreadscaler
          properties:
            replicas: 1
```

It would be important here to note the difference between provider configuration and the linkdef trait, since they can be used for different things.

stephenandary:
I can take this on since it sounded easy

stephenandary:
Files to keep in mind when I start this Monday: 

`models/mod.rs`,`models/internals.rs`, and `server/parser.rs`

PR #109, opened 2023-05-08
Support config for provider start commands

stephenandary:
## Feature or Problem
Be able to pass config to a starting provider that is an alternative to a linkdef which would be desired if the same config was to be used with all linkdefs. 

## Related Issues
(https://github.com/wasmCloud/wadm/issues/102)

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows


src/model/mod.rs
```diff
@@ -114,6 +114,9 @@ pub struct CapabilityProperties {
     pub image: String,
     /// The contract ID of this capability
     pub contract: String,
+    /// An opaque config string which can be passed to a stating provider. See wasmcloud_interface_lattice_control/struct.StartProviderCommand for more details
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub config: Option<String>,
```
brooksmtownsend, line 119: @stephenandary I think I wrote that part up in the PR, and I can't remember when we discussed, but I think it would be very nice if this `config` variable could be just a YAML list of String -> String. e.g. instead of 
```
config: {"address": "0.0.0.0:8080"}
```
We could have native support for
```yaml
config:
  address: 0.0.0.0:8080
```

We would effectively take that list of config, turn it into JSON, and then fire it off with the command. Thoughts?

stephenandary: I reached out to Taylor to ask if the purpose of this comment:
```
    // TODO: Do we need to support config_json paths?
    /// Additional annotations to attach on this command
    pub annotations: HashMap<String, String>,
```
was for that purpose, and if it was roughed in already because it was not clear in the [issue](https://github.com/wasmCloud/wadm/issues/102) or from the interface what you want done relative to various providers. 

The issue states:  "It would be nicer to specify nested config as YAML or JSON instead of just an opaque string, but the provider itself may or may not support a specific format" 

So I assumed that what the issue was asking for was a base64 encoded opaque string that the user creates as that was accepted previously in wadm. 

Is the ask for the manifest to be parsed, and those fields to be converted to an opaque string of JSON, YAML, or whatever format is accepted by the provider? 

If so, maybe it would be nice to define config format as another field then so the user can indicate what the provider is accepting and set it up for JSON, YAML, or Other MIME type. 

If passing as Other, maybe it can be just urlencoded or an array or something.

stephenandary: I am assuming that is what the HashMap is for, and you just want to parse the config into that, but I am not sure yet

thomastaylor312: @brooksmtownsend I know we talked about this, but I did double check that we do not put a requirement on the data type, although currently every provider we know of expects JSON. If we want to change that, we should probably RFC that to the provider specification.

However, with that said @stephenandary, I know @brooksmtownsend and I talked about an alternate solution here: Have both a `config_string` and `config_json` field available for use (with `config_string` taking precedence). We know that JSON is the most common format here, so offering that as a short cut for freeform data that looks like the below would be nice:

```yaml
properties:
    config_json:
        max_conns: 25
        default_address: 0.0.0.0:8080
```

With the config_string being the easy fallback.

@brooksmtownsend, thoughts here?

thomastaylor312: Actually, as soon as I sent that, I realized that would be a bad idea. Then everyone is going to want their config format type supported (i.e. `config_yaml`, `config_toml`). So pending @brooksmtownsend's thoughts, I think we just keep this as `config` and have it be a base64 encoded string

stephenandary: That is what I was trying to wrap my head around when I mentioned above it may be nice to define an enum of MIME types that are supported, and you can match on that to encode based on a field on manifest to indicate type. 

e.g. 
```
properties:
    config:
        format: json
        max_conns: 25
        default_address: 0.0.0.0:8080
````
Then if you wanted to support more MIME types it would just be a matter of adding an enum variant and creating some additional impls or an Into Trait that could potentially just convert a hashmap defined on manifest to the defined format?

Comment (stephenandary) [reply src/model/mod.rs:119]: If a format was not supported maybe then you could default to hashmap of opaque strings or something

Comment (brooksmtownsend) [reply src/model/mod.rs:119]: So right now, the config is actually expected to be JSON and supplying anything else wouldn't really work for the start provider command, see https://docs.rs/wasmbus-rpc/0.13.0/wasmbus_rpc/core/struct.HostData.html#structfield.config_json

Of course we don't _verify_ that the string is JSON, but it's probably a bad idea to send a regular string there. I think for now we should probably encode it as a JSON string, since there's not really a way to send anything else

Comment (brooksmtownsend) [reply src/model/mod.rs:119]: The RFC that would be nice would be something like sending config as a b64 encoded string and then unpacking that string into the data structure that the provider expects

File: src/model/mod.rs [hunk 2]
```diff
@@ -114,6 +114,9 @@ pub struct CapabilityProperties {
     pub image: String,
     /// The contract ID of this capability
     pub contract: String,
+    /// Optional provider config
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub config: Option<serde_json::Value>,
```
stephenandary, line 119: the assumption is that serde_json::Value here will be String variant, so I am using Option<String> elsewhere, but I need to look into how this works.

stephenandary: Think about how yaml syntax of  something like 
```
config:
    address: 0.0.0.0
```
will be processed here

PR #112, opened 2023-05-09
feat(*) Remove jitter with plural events and expected events

brooksmtownsend:
There's one or two more TODOs that I've outlined below, but I'm marking as ready for review to get some input around the rest of the methodology. 

This PR:
- [x] Adds support for ActorsStarted / ActorsStopped and their failure events
- [x] Implicitly requires `wadm` to work with v0.63+ versions of the host because of the above
- [x] Removes the notion of a backoff, replacing it with logic to make Scalers compute "expected events" for each command that they send. For example, if a scaler wants to start 5 of an actor, it won't take any action until it either receives an event where 5 of that actor started, or if that command failed to complete. Scaler implementers do not need to care about this, it's all on the `wadm` side.
- [x] TODO: Make the expected events list clear after a certain time frame to account for missed events or strange edge cases.

## Feature or Problem
Currently, Scalers (particularly the ActorSpreadScaler) are prone to jitter and difficulty actually converging on proper state when you run many (10+) copies of an actor. They are handling each event as it comes in, and we added a very coarse backoff mechanism to account for this but it still creates difficulties.

## Related Issues
Will file issue soon to make state only update on plural events.

## Release Information
next alpha release

## Consumer Impact
Consumers will notice that wadm feels more responsive at reconciling

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
This feature is large enough that it probably warrants some tests to verify behavior around expected events.

- [ ] TODO: Tests

### Manual Verification
I've applied applications and messed with them in many ways to confirm that this works great with 1 running wadm.

### DIFF
--- a/.devcontainer/install.sh
+++ b/.devcontainer/install.sh
@@ -1,20 +1,5 @@
 #!/bin/bash
 
 # INSTALL WASH
-cargo install wash-cli --git https://github.com/wasmcloud/wash --branch feat/wadm_0.4_support --force 
-
-# INSTALL WADM
-ARCH=$(arch)
-VERSION=v0.4.0-alpha.1
-
-if [[ $ARCH == "x86_64" ]]; then
-    ARCH="amd64"
-fi
-
-TARBALL=wadm-$VERSION-linux-$ARCH
-
-curl -fLO https://github.com/wasmCloud/wadm/releases/download/$VERSION/$TARBALL.tar.gz
-tar -xvf $TARBALL.tar.gz
-chmod +x $TARBALL/wadm
-mv $TARBALL/wadm /usr/local/cargo/bin/wadm
-rm -rf $TARBALL $TARBALL.tar.gz
+curl -s https://packagecloud.io/install/repositories/wasmcloud/core/script.deb.sh | sudo bash
+sudo apt install wash openssl -y
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -9,8 +9,8 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     events::{
-        ActorClaims, ActorsStartFailed, ActorsStarted, Event, Linkdef, LinkdefSet, ProviderClaims,
-        ProviderStartFailed, ProviderStarted,
+        ActorClaims, ActorsStartFailed, ActorsStarted, ActorsStopped, Event, Linkdef, LinkdefSet,
+        ProviderClaims, ProviderStartFailed, ProviderStarted,
     },
     model::CapabilityConfig,
     workers::insert_managed_annotations,
@@ -46,7 +46,7 @@ impl Command {
     /// # Return
     /// - The first element in the tuple corresponds to the "success" event a host would output after completing this command
     /// - The second element in the tuple corresponds to an optional "failure" event that a host could output if processing fails
-    pub fn corresponding_event(&self, model_name: &str) -> (Event, Option<Event>) {
+    pub fn corresponding_event(&self, model_name: &str) -> Option<(Event, Option<Event>)> {
         match self {
             Command::StartActor(StartActor {
                 annotations,
@@ -57,7 +57,7 @@ impl Command {
             }) => {
                 let mut annotations = annotations.to_owned();
                 insert_managed_annotations(&mut annotations, model_name);
-                (
+                Some((
                     Event::ActorsStarted(ActorsStarted {
                         annotations: annotations.to_owned(),
                         image_ref: reference.to_owned(),
@@ -75,7 +75,28 @@ impl Command {
                         // We do not know the public key from the command
                         public_key: String::with_capacity(0),
                     })),
-                )
+                ))
+            }
+            Command::StopActor(StopActor {
+                annotations,
+                actor_id,
+                host_id,
+                count,
+                ..
+            }) => {
+                let mut annotations = annotations.to_owned();
+                insert_managed_annotations(&mut annotations, model_name);
+                Some((
+                    Event::ActorsStopped(ActorsStopped {
+                        annotations: annotations.to_owned(),
+                        public_key: actor_id.to_owned(),
+                        host_id: host_id.to_owned(),
+                        count: *count,
+                        // We don't know if there are additional instances, so can't consider remaining
+                        remaining: 0,
+                    }),
+                    None,
+                ))
             }
             Command::StartProvider(StartProvider {
                 annotations,
@@ -86,7 +107,7 @@ impl Command {
             }) => {
                 let mut annotations = annotations.to_owned();
                 insert_managed_annotations(&mut annotations, model_name);
-                (
+                Some((
                     Event::ProviderStarted(ProviderStarted {
                         annotations: annotations.to_owned(),
                         claims: ProviderClaims::default(),
@@ -109,7 +130,7 @@ impl Command {
                         // We don't know this field from the command
                         error: String::with_capacity(0),
                     })),
-                )
+                ))
             }
             Command::PutLinkdef(PutLinkdef {
                 actor_id,
@@ -118,7 +139,7 @@ impl Command {
                 contract_id,
                 values,
                 ..
-            }) => (
+            }) => Some((
                 Event::LinkdefSet(LinkdefSet {
                     linkdef: Linkdef {
                         actor_id: actor_id.to_owned(),
@@ -131,8 +152,8 @@ impl Command {
                     },
                 }),
                 None,
-            ),
-            _ => todo!(),
+            )),
+            _ => None,
         }
     }
 }
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -7,7 +7,14 @@ use std::{
 
 use serde::{Deserialize, Serialize};
 
-use crate::model::CapabilityConfig;
+use crate::{
+    events::{
+        ActorClaims, ActorsStartFailed, ActorsStarted, Event, Linkdef, LinkdefSet, ProviderClaims,
+        ProviderStartFailed, ProviderStarted,
+    },
+    model::CapabilityConfig,
+    workers::insert_managed_annotations,
+};
 
 macro_rules! from_impl {
     ($t:ident) => {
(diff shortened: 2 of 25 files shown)


src/scaler/spreadscaler/mod.rs
```diff
@@ -62,15 +61,14 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ActorSpreadScaler<S> {
         };
         self.config.spread_config = spread_config;
         self.spread_requirements = compute_spread(&self.config.spread_config);
+        //TODO: do we need to remove the backoff?
```
thomastaylor312, line 64: I assume yes at this point? 😄

src/workers/event.rs
```diff
@@ -701,53 +701,90 @@ where
...
+///
+/// # Returns
+///
+/// * `Result<Vec<Command>>`: A `Result` containing a vector of `Command` structs if successful,
+///   or an error of type `anyhow::Error` if any error occurs while processing the event.
+async fn scaler_handle_event(scaler: &ScalerWithEvents, event: &Event, model_name: &str) -> anyhow::Result<Vec<Command>> {
+    // println!("Handling event {:?}", event);
+
+    // If a scaler is expecting events still, don't have it handle events
+    //TODO: we may need to expire this after a time period.
```
thomastaylor312, line 761: Yep, this should clear out after a while

thomastaylor312 (commented):
Some initial comments. Might want to schedule a synchronous review on this one with you

src/commands/mod.rs
```diff
@@ -7,6 +7,14 @@ use std::{
 
 use serde::{Deserialize, Serialize};
 
+use crate::{
+    events::{
+        ActorClaims, ActorsStartFailed, ActorsStarted, Event, Linkdef, LinkdefSet, ProviderClaims,
+        ProviderStartFailed, ProviderStarted,
+    },
+    workers::insert_managed_annotations,
```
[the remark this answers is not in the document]
brooksmtownsend: Moved to event_helpers in workers, I actually don't mind it there. Would be fine considering moving though if you have another suggestion

.devcontainer/install.sh
```diff
@@ -1,20 +1,5 @@
...
-TARBALL=wadm-$VERSION-linux-$ARCH
-
-curl -fLO https://github.com/wasmCloud/wadm/releases/download/$VERSION/$TARBALL.tar.gz
-tar -xvf $TARBALL.tar.gz
-chmod +x $TARBALL/wadm
-mv $TARBALL/wadm /usr/local/cargo/bin/wadm
-rm -rf $TARBALL $TARBALL.tar.gz
\ No newline at end of file
+curl -s https://packagecloud.io/install/repositories/wasmcloud/core/script.deb.sh | sudo bash
+sudo apt install wash openssl -y
```
brooksmtownsend, line 5: Note: devcontainer only takes like 30 seconds to startup now!

brooksmtownsend:
Pretty positive I handled the rebase properly, but if I removed something from #118 lmk

src/events/types.rs
```diff
@@ -83,10 +83,14 @@ pub trait EventType {
 }
 
 /// A lattice event
-#[derive(Debug)]
+// TODO:  this `Deserialize` impl _compiles_ but doesn't let you deserialize an event because of our custom Serialize
```
brooksmtownsend, line 86: @thomastaylor312 just a note, this is probably why tests fail and all. The notifications that we try to send to remove an event or register events fail to deserialize on the other end because they use the derived `Deserialize` but the manually implemented `Serialize`

src/commands/mod.rs
```diff
@@ -30,6 +37,127 @@ pub enum Command {
...
+                        public_key: actor_id.to_owned(),
+                        host_id: host_id.to_owned(),
+                        count: *count,
+                        // We don't know if there are additional instances, so can't consider remaining
+                        remaining: 0,
+                    }),
+                    None,
+                ))
+            }
+            Command::StartProvider(StartProvider {
```
connorsmith256, line 101: No `StopProvider` or `DeleteLinkdef`?

src/events/types.rs
connorsmith256: @thomastaylor312 is this still an issue?

src/scaler/manager.rs
```diff
@@ -32,108 +31,77 @@ use crate::{
...
 
 /// All events sent for manifest notifications
 #[derive(Debug, Serialize, Deserialize)]
 pub enum Notifications {
     CreateScalers(Manifest),
     DeleteScalers(String),
-    BackoffScaler(String),
+    /// Register expected events for a manifest. You can either trigger this with an event (which
+    /// will result in calling `handle_event`) or without in order to calculate expected events with
+    /// a full reconcile. is being done (like on first deploy), rather than just handling a single
```
connorsmith256, line 54: > is being done

Might have deleted something?

src/scaler/manager.rs
```diff
@@ -298,32 +262,35 @@ where
...
+    pub async fn get_specific_scaler<'a>(
+        &'a self,
+        name: &'a str,
+        scaler_id: &'a str,
+    ) -> Option<SingleScaler> {
+        let lock = self.scalers.clone().read_owned().await;
+        OwnedRwLockReadGuard::try_map(lock, |scalers| match scalers.get(name) {
+            Some(scalers) => scalers.iter().find(|s| s.id() == scaler_id),
+            None => None,
+        })
```
connorsmith256, line 284: Supernit: does `and_then` work?
```rust
scalers
    .get(name)
    .and_then(|scalers| scalers.iter().find(|s| s.id() == scaler_id)
```

src/scaler/manager.rs
```diff
@@ -441,16 +406,43 @@ where
...
+                                                Ok(e) => e,
+                                                Err(e) => {
+                                                    error!(error = %e, %name, "Unable to parse given event");
+                                                    continue;
+                                                }
+                                            };
+                                            if let Err(e) = scaler.handle_event(&parsed_event).await {
+                                                error!(error = %e, %name, %scaler_id, "Unable to register expected events for scaler");
+                                            }
+                                        } else if let Err(e) = scaler.reconcile().await {
```
connorsmith256, line 424: Not sure I follow this. Why do we call `reconcile` if there's no triggering event? (And why do we not call reconcile when removing an expected event?)

src/scaler/mod.rs
```diff
@@ -39,18 +60,363 @@ pub trait Scaler {
...
+    async fn remove_event(&self, event: &Event) -> Result<bool> {
+        let mut expected_events = self.expected_events.write().await;
+        let before_count = expected_events.len();
+        expected_events.retain(|(success, fail)| {
+            !evt_matches_expected(success, event)
+                && !fail
+                    .as_ref()
+                    .map(|f| evt_matches_expected(f, event))
+                    .unwrap_or(false)
+        });
```
connorsmith256, line 141: This is hard to follow. Are we removing if the event matches _either_ the success or fail in any of the `expected_events`?

src/scaler/mod.rs
```diff
@@ -39,18 +60,363 @@ pub trait Scaler {
...
+            trace!("Scaler is not backing off, handling event");
+            let commands = self.scaler.handle_event(event).await?;
+
+            // Based on the commands, compute the events that we expect to see for this scaler. The scaler
+            // will then ignore incoming events until all of the expected events have been received.
+            let expected_events = commands
+                .iter()
+                .filter_map(|cmd| cmd.corresponding_event(model_name));
+
+            // Only let other scalers know if we generated commands to take
```
connorsmith256, line 192: > if we generated commands to take

To take what?

src/scaler/mod.rs
```diff
@@ -39,18 +60,363 @@ pub trait Scaler {
...
+            // NOTE(brooksmtownsend): I tried to make the events iterator peekable and look at that instead
+            // but that resulted in a "higher ranked lifetime error". Weird.
```
connorsmith256, line 194: Could link to the rust issue with a TODO to revisit once that's fixed?

src/scaler/mod.rs
```diff
@@ -39,18 +60,363 @@ pub trait Scaler {
...
+        // Clear any existing handle
+        if let Some(handle) = event_cleaner.take() {
+            handle.abort();
+        }
+        let expected_events = self.expected_events.clone();
+        let timeout = self.cleanup_timeout;
+
+        *event_cleaner = Some(tokio::spawn(async move {
+            tokio::time::sleep(timeout).await;
+            // We don't use the method to clear here because of borrowing semantics.
```
connorsmith256, line 263: I don't understand this comment, since we are calling `clear` on the next line

src/scaler/mod.rs
```diff
@@ -39,18 +60,363 @@ pub trait Scaler {
...
+            }
+            Ok(commands) => {
+                trace!("Reconcile generated no commands, no need to register expected events");
+                Ok(commands)
+            }
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Sets a timed cleanup task to clear the expected events list after a timeout
```
connorsmith256, line 251: Why is this needed?

src/workers/event.rs
```diff
@@ -853,6 +864,39 @@ where
...
+    Fut: futures::Future<Output = Result<Vec<Command>>>,
+    I: IntoIterator<Item = Fut>,
+{
+    let (commands, errors): (Vec<_>, Vec<_>) = futures::future::join_all(futs)
+        .await
+        .into_iter()
+        .partition(Result::is_ok);
+    // SAFETY: We partitioned by Ok and Err, so the unwraps are fine
+    (
+        commands.into_iter().flat_map(Result::unwrap).collect(),
```
connorsmith256, line 880: I assume this is right, but having a hard time parsing why it's `flat_map` instead of `map`

connorsmith256 (commented):
Mostly questions on my part

src/events/types.rs
thomastaylor312: Nope, all fixed now, forgot to remove todo

src/commands/mod.rs
thomastaylor312: I'm not sure what @brooksmtownsend was going for here, but my guess is that both of those don't matter for backing off. If the scaler was the one that stopped a provider, that is fine as it will get picked up by the scaler, see that it doesn't need to reconcile, and continue on. This is due to only one provider being able to run per link name + host combo.

The linkdef is similar, if we want to delete it, then it doesn't matter when we receive it as there can only ever be one linkdef (per link, provider, actor).

src/scaler/manager.rs
thomastaylor312: Why yes it does

src/scaler/manager.rs
thomastaylor312: This is cheating and I'll add a comment to clarify. Basically, if this is a backoff wrapped scaler, it will cause the expected events to be added to the list. If it isn't we just return the commands and do nothing with them

src/scaler/mod.rs
thomastaylor312: Yes. Any command we send out could have either a success case or an optional error case. Matching either one means we found one of the expected events

src/scaler/mod.rs
thomastaylor312: We don't want to stay in a backoff state forever, this handles the case of a network interruption or missed events by eventually expiring them

src/workers/event.rs
thomastaylor312: The inner type of each result is `Vec<Command>`, so this shortcuts it from `commands.into_iter().map(Result::unwrap).flatten().collect()`. I think it was clippy who told me not to do that

src/scaler/mod.rs
thomastaylor312: I think it was left over from something before the refactor

src/scaler/mod.rs
thomastaylor312: I just removed this for now (cc @brooksmtownsend), I wasn't sure what he was trying to do with peekable. Can add back in later

src/scaler/mod.rs
connorsmith256: It took me a minute to grok what was happening. Mind adding a comment there?

connorsmith256 (approved):

PR #116, opened 2023-05-17
ref(server): Refactors storage for manifests 

thomastaylor312:
This solved the problem where you could really only update one manifest at a time per lattice. Please note that this has some added e2e tests, but those are currently commented out as they are waiting on some of our other work for removing jitter from starting actors (which is landing in the new host version and in wadm fairly soon)

### DIFF
--- a/src/scaler/manager.rs
+++ b/src/scaler/manager.rs
@@ -5,6 +5,7 @@ use std::{collections::HashMap, ops::Deref, sync::Arc};
 use anyhow::Result;
 use async_nats::jetstream::{
     consumer::pull::{Config as PullConfig, Stream as MessageStream},
+    kv::Store as KvStore,
     stream::Stream as JsStream,
     AckKind,
 };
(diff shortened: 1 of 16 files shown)


src/scaler/manager.rs
```diff
@@ -198,11 +196,23 @@ where
             .map_err(|e| anyhow::anyhow!("Unable to subscribe to consumer: {e:?}"))?;
 
         // Get current scalers set up
-        let all_manifests = manifest_store.list::<StoredManifest>(lattice_id).await?;
+        let manifest_store = crate::server::ModelStorage::new(manifest_store);
+        let futs = manifest_store
```
brooksmtownsend, line 200: 🦶🏻

src/server/storage.rs
```diff
@@ -0,0 +1,234 @@
...
+    #[instrument(level = "debug", skip(self))]
+    pub async fn list(&self, lattice_id: &str) -> Result<Vec<ModelSummary>> {
+        debug!("Fetching list of models from storage");
+        let futs = self
+            .get_model_set(lattice_id)
+            .await?
+            .unwrap_or_default()
+            .0
+            .into_iter()
+            // We can't use filter map with futures, but we can use map and then filter_map it out below
```
brooksmtownsend, line 97: Did you want to add this `filter_map` or was it not necessary in the end?

src/server/types.rs
```diff
@@ -57,7 +57,7 @@ pub struct ModelSummary {
     pub name: String,
     pub version: String,
     pub description: Option<String>,
-    pub deployed: bool,
+    pub deployed_version: Option<String>,
```
brooksmtownsend, line 60: :alert: breaking change, this will need to be updated in `wash` as well

brooksmtownsend (approved):

src/server/storage.rs
thomastaylor312: It wasn't necessary. I'll remove the comment

src/server/types.rs
thomastaylor312: I think this is more accurate and useful though! Do you?

PR #117, opened 2023-05-19
feat(*) support provider config json

### LINKED ISSUE #102: [FEATURE] Support `config_json` for provider start commands

brooksmtownsend:
Still draft as it should be merged after #112. Can always break it out into its own PR if needed but I wanted to test this in tandem with the other change. 

## Feature or Problem
This PR adds support for a `config` block in the provider yaml. For example, you can change the default URL of the wasmCloud redis provider like this so you don't have to specify a URL in the linkdef
```yaml
    - name: redis
      type: capability
      properties:
        image: wasmcloud.azurecr.io/kvredis:0.19.0
        contract: wasmcloud:keyvalue
        config:
          url: redis://127.0.0.1:6380
```

I started with the implementation of this as a `HashMap<String, String>` even though it could technically be better represented as a `String, serde_json::Value`, just to keep it simple. It all ends up being serialized into a String anyways, but I pushed an additional commit with that change so if you specified a port like `8080`, it actually gets serialized more accurately as a number. Need to consider if this is better or more complicated, from my testing it should deserialize properly when users expect certain types.

## Related Issues
Fixes #102 

## Release Information
next alpha

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Updated unit tests to set config to be `None`

### Manual Verification
I ran this manually with the kvcounter yaml snippet above, which properly made it to the provider

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -223,9 +223,10 @@ pub struct StartProvider {
     pub link_name: Option<String>,
     /// The name of the model/manifest that generated this command
     pub model_name: String,
-    // TODO: Do we need to support config_json paths?
     /// Additional annotations to attach on this command
     pub annotations: HashMap<String, String>,
+    /// Optional provider configuration to pass to the provider
+    pub config_json: Option<String>,
 }
 
 from_impl!(StartProvider);
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -117,6 +117,9 @@ pub struct CapabilityProperties {
     /// An optional link name to use for this capability
     #[serde(skip_serializing_if = "Option::is_none")]
     pub link_name: Option<String>,
+    /// Optional configuration for this capability provider
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub config: Option<HashMap<String, String>>,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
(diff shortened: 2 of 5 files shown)


brooksmtownsend:
Closing since @thomastaylor312 is cooler than me in #118

PR #118, opened 2023-05-19
feat(*): Adds support for provider config

### LINKED ISSUE #102: [FEATURE] Support `config_json` for provider start commands

thomastaylor312:
Adds support for passing a string encoded config or an optional json encoded config through to a provider. This includes an additional example manifest and tests

Closes #102

### DIFF
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -1,6 +1,10 @@
 use std::collections::{BTreeMap, HashMap};
 
-use serde::{Deserialize, Serialize};
+use serde::{
+    de::{self, Deserializer, MapAccess, Visitor},
+    ser::Serializer,
+    Deserialize, Serialize,
+};
 
 pub(crate) mod internal;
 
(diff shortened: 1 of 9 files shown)


src/model/mod.rs
```diff
@@ -117,6 +121,64 @@ pub struct CapabilityProperties {
...
+    }
+}
+
+struct CapabilityConfigVisitor;
+
+impl<'de> Visitor<'de> for CapabilityConfigVisitor {
+    type Value = CapabilityConfig;
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+        formatter.write_str("string or json")
```
brooksmtownsend, line 153: Should we clarify here that json == json or YAML?

src/server/handlers.rs
```diff
@@ -47,6 +50,31 @@ impl<P: Publisher> Handler<P> {
...
+            if let Properties::Capability {
+                properties:
+                    CapabilityProperties {
+                        config: Some(CapabilityConfig::Json(data)),
+                        ..
+                    },
+            } = &component.properties
+            {
+                if let Err(e) = serde_json::to_string(data) {
+                    self.send_error(
```
brooksmtownsend, line 65: 👨🏻‍🍳 💋

brooksmtownsend (approved):
Love it.

src/model/mod.rs
thomastaylor312: Good idea

PR #128, opened 2023-06-05
Adds support for building arm image

### LINKED ISSUE #127: Streamline docker image build and add aarch64 target
We should probably follow a similar process to the OTP pipeline and just download the artifacts and then build for both x86 and aarch64 simultaneously. Right now we are building wadm again in the container and only for x86

thomastaylor312:
Also makes image building faster by copying the already built binary

Closes #127

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -195,6 +195,18 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
+      - uses: actions/download-artifact@v3
+        with:
+          name: wadm-${{ env.RELEASE_VERSION }}-linux-aarch64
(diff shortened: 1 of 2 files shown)


thomastaylor312:
Note that I am letting this run an example build before returning some of the `if` blocks to their original form

thomastaylor312:
Ok, this is ready for review. You can see a test build with this here: https://github.com/wasmCloud/wadm/pkgs/container/wadm/98978140?tag=taylor-test

connorsmith256 (approved):

brooksmtownsend (approved):
Only downside to this way of doing the docker build is that you can't just build the image yourself unless you can compile the linux binary, but that's ok for now and I'd rather have that as a good-first-issue than block

PR #132, opened 2023-07-11
feat(multitenant)!: support multitenant lattice orchestration

brooksmtownsend:
## Feature or Problem
This PR fully implements the `--multitenant` feature which allows wadm to manage multiple lattices, where events come in with an account prefix and control interface commands are sent out with the same prefix. The [configuring wadm](https://wasmcloud.com/docs/fundamentals/wadm/deploying#configuring-credentials) documentation shows how to create credentials for a basic scenario. This PR also includes e2e tests

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

PR #133, opened 2023-07-18
feat(multitenant)! Account prefixed API topics and manifest storage

brooksmtownsend:
## Feature or Problem
This PR ensures we are storing and querying manifests based on an account prefix if running in multitenant mode. This ensures that accounts can only access and modify manifests that are specifically bound to that account, rather than simply using a global lattice ID.

Due to possible edge cases, it's still recommended to keep lattice prefixes globally unique between accounts, as there are not e2e tests at this time to verify that functionality works without issues. This PR at least implements the portion of that that constrains manifests from conflicting between accounts.

## Related Issues
#132

## Release Information
v0.5.0

## Consumer Impact
Multitenancy hasn't released yet so this shouldn't affect consumers, however they will need to examine the account prefixed topic if manually looking up manifests in NATS.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Modified a few unit tests that were hitting deprecation warnings

### Acceptance or Integration
Will have to modify the e2e tests to properly import/export account prefixes

### Manual Verification
Using the multitenant docker compose, I was able to verify that if I put a manifest using one account, I cannot query that manifest in another account even if I use the correct lattice prefix 😄

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -26,6 +26,7 @@ pub struct Server<P> {
     handler: Handler<P>,
     subscriber: Subscriber,
     prefix: String,
+    multitenant: bool,
 }
 
 impl<P: Publisher> Server<P> {
@@ -39,6 +40,7 @@ impl<P: Publisher> Server<P> {
         store: Store,
         client: Client,
         topic_prefix: Option<&str>,
+        multitenant: bool,
         notifier: ManifestNotifier<P>,
     ) -> anyhow::Result<Server<P>> {
         // Trim off any spaces or trailing/preceding dots
@@ -51,7 +53,13 @@ impl<P: Publisher> Server<P> {
             anyhow::bail!("Given prefix was empty")
         }
 
-        let topic = format!("{prefix}.>");
+        let topic_prefix = if multitenant {
+            format!("*.{prefix}")
+        } else {
+            prefix.clone()
+        };
+
+        let topic = format!("{topic_prefix}.>");
         info!(%topic, "Creating API subscriber");
         // NOTE(thomastaylor312): Technically there is a condition where two people try to send an
         // update to the same manifest. We are protected against this overwriting each other (we
@@ -71,6 +79,7 @@ impl<P: Publisher> Server<P> {
             },
             subscriber,
             prefix,
+            multitenant,
         })
     }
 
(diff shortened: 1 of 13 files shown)


src/server/mod.rs
```diff
@@ -51,7 +53,15 @@ impl<P: Publisher> Server<P> {
...
 
-        let topic = format!("{prefix}.>");
+        let topic_prefix = if multitenant {
+            format!("*.{prefix}")
+        } else {
+            prefix.clone()
+        };
+
+        let topic = format!("{topic_prefix}.>");
+        eprintln!("GONNA LISTEN ON: {topic}");
```
connorsmith256, line 63: Oopsie whoopsie

src/server/mod.rs
```diff
@@ -81,7 +92,7 @@ impl<P: Publisher> Server<P> {
     #[instrument(level = "info", skip_all)]
     pub async fn serve(mut self) -> anyhow::Result<()> {
         while let Some(msg) = self.subscriber.next().await {
-            if !msg.subject.starts_with(&self.prefix) {
+            if !msg.subject.starts_with(&self.prefix) && !self.multitenant {
                 warn!(subject = %msg.subject, "Received message on an invalid subject");
```
connorsmith256, line 96: When would this happen anyway? I assume the subscription was always on `{prefix}.>` before, and now can also be `*.{prefix}.>`

src/server/mod.rs
```diff
@@ -196,6 +252,7 @@ impl<P: Publisher> Server<P> {
             anyhow::bail!("Found extra components of subject")
         }
         Ok(ParsedSubject {
+            account_id: account_id,
```
connorsmith256, line 255: Nit: just `account_id`?

connorsmith256 (approved):

src/server/mod.rs
brooksmtownsend: I had full confidence and definitely never needed to verify this

connorsmith256 (approved):

PR #136, opened 2023-07-24
fix(*): support deleting all versions of manifest

### LINKED ISSUE #129: [BUG] cannot recreate app name after all its versions have been deleted
# Summary:
When you delete the last or only version of an app it seems to "burn" that app name such that all subsequent `wash app put` calls fail with `Internal storage error`

# Logs from `wadm`:
```
ERROR serve: wadm::server::handlers: Unable to store updated data error=Custom { kind: Other, error: "nats: error while pub       │ lishing message: 10071, 400, wrong last sequence: 8" }
```

# Steps to reproduce:
```
# cleanup local env
wash down --all && rm -rf $HOME/.wash && rm -rf /tmp/nats/*

# write simple echo.yml
cat <<EOF > echo.yml
kind: Application
metadata:
  name: echo
  annotations:
    version: v0.0.1
    description: "This is my app"
spec:
  components:
    - name: echo
      type: actor
      properties:
        image: wasmcloud.azurecr.io/echo:0.3.7
      traits:
        - type: spreadscaler
          properties:
            replicas: 1
EOF

# and repro
wash up -d && wash app put ./echo.yml && wash app del echo v0.0.1 && wash app put ./echo.yml
```

brooksmtownsend:
## Feature or Problem
This PR fixes two issues
1. As outlined in #129, we had issues with fully deleting all versions of a manifest and then putting that manifest again, this is both fixed by my filter to ensure we only `update` if revision > 0 and adding the jetstream domain to the purge operation
2. For a non-multitenant wadm and multitenant wadm that share the same NATS infrastructure, avoid creating consumers that will automatically fail.

## Related Issues
Fixes #129 
Depends on https://github.com/nats-io/nats.rs/pull/1055

## Release Information
v0.5.0

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/src/consumers/manager.rs
+++ b/src/consumers/manager.rs
@@ -111,6 +111,7 @@ impl<C> ConsumerManager<C> {
         permit_pool: Arc<Semaphore>,
         stream: NatsStream,
         worker_generator: F,
+        multitenant: bool,
     ) -> ConsumerManager<C>
     where
         W: Worker + Send + Sync + 'static,
@@ -150,6 +151,13 @@ impl<C> ConsumerManager<C> {
                     (Some(id), prefix) => (id, prefix),
                     (None, _) => return None,
                 };
+
+                // Don't create multitenant consumers if running in single tenant mode, and vice versa
+                if multitenant_prefix.is_some() != multitenant {
+                    trace!(%lattice_id, "Skipping consumer for lattice because multitenant doesn't match");
+                    return None;
+                }
+
                 // NOTE(thomastaylor312): It might be nicer for logs if we add an extra param for a
                 // friendly consumer manager name
                 trace!(%lattice_id, subject = %info.config.filter_subject, "Adding consumer for lattice");
--- a/src/server/storage.rs
+++ b/src/server/storage.rs
@@ -74,7 +74,7 @@ impl ModelStorage {
         let key = model_key(account_id, lattice_id, model.name());
         trace!(%key, "Storing manifest at key");
         let data = serde_json::to_vec(&model).map_err(anyhow::Error::from)?;
-        if let Some(revision) = current_revision {
+        if let Some(revision) = current_revision.filter(|r| r > &0) {
             self.store
                 .update(&key, data.into(), revision)
                 .await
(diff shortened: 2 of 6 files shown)


src/consumers/manager.rs
```diff
@@ -150,6 +151,13 @@ impl<C> ConsumerManager<C> {
                     (Some(id), prefix) => (id, prefix),
                     (None, _) => return None,
                 };
+
+                // Don't create multitenant consumers if running in single tenant mode, and vice versa
+                if multitenant_prefix.is_some() != multitenant {
+                    trace!(%lattice_id, "Skipping consumer for lattice because multitenant doesn't match");
+                    return None;
+                }
```
connorsmith256, line 159: Is this expected sometimes? It feels like a caller error. Should it be a `warn` instead?

src/server/storage.rs
```diff
@@ -74,7 +74,7 @@ impl ModelStorage {
         let key = model_key(account_id, lattice_id, model.name());
         trace!(%key, "Storing manifest at key");
         let data = serde_json::to_vec(&model).map_err(anyhow::Error::from)?;
-        if let Some(revision) = current_revision {
+        if let Some(revision) = current_revision.filter(|r| r > &0) {
```
connorsmith256, line 77: I don't think I understand this. Is the initial revision for a model always 0?

src/consumers/manager.rs
brooksmtownsend: This could happen often if a multitenant wadm and a non-multitenant wadm are using the same streams, since there will be consumers for both. Perhaps a debug, but this isn't an issue as long as you are running in the mode that you are supposed to

src/server/storage.rs
brooksmtownsend: Effectively here, having a revision of `0` and having a `None` revision should have the same outcome. So the initial revision is `0`, and we don't want to use `update` there we want to use `put` to create the model in the first place. This was just fixing an issue where initial create would have `None`, but after deleting a model the revision would come back from the entry API as `Some(0)` and would fail the update.

connorsmith256 (approved):

PR #139, opened 2023-07-25
feat(*): support non-string spreadscaler requirements

### LINKED ISSUE #131: [BUG] wadm spreadscaler requirements don't work as numbers
I defined a spreadscaler requirement as a number, like below.
```yaml
        spread:
          - name: myspread
            weight: 100
            requirements:
              host_number: 0
```

Unfortunately, it seems we can parse that number as a JSON value that's not a string, so even though my host had the label `host_number=0` it did not match the scaler property. The fix was as follows:
```yaml
        spread:
          - name: myspread
            weight: 100
            requirements:
              host_number: "0"
```

Because host labels are always interpreted as string:string maps, we should either enforce that in our parsing of the manifest or parse host labels as JSON values. I think the better solution is to keep it simple and parse spread key/value requirements as strings, but am open to other opinions as well.

brooksmtownsend:
Signed-off-by: Brooks Townsend <brooks@cosmonic.com>

## Feature or Problem
This PR allows spreadscaler requirements to be parsed as a string->value map, instead of just string->string. This lets you supply boolean and numeric values for spreadscaler properties and have them parse properly instead of falling under the custom object. Additionally, it modifies the `eligible_hosts` helper function to properly compare the `serde_json::Value` to the host labels that are still string->string hashmaps

## Related Issues
Fixes #131

## Release Information
0.5.0

## Consumer Impact
This may cause previously "custom" scaler objects to be parsed as spreadscaler objects, but I find it unlikely that it would solely be based on this change.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Added 2 unit tests to ensure parsing works properly and that matching based on labels works as intended.

### DIFF
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -266,7 +266,7 @@ pub struct Spread {
     pub name: String,
     /// An arbitrary map of labels to match on for scaling requirements
     #[serde(skip_serializing_if = "BTreeMap::is_empty")]
-    pub requirements: BTreeMap<String, String>,
+    pub requirements: BTreeMap<String, serde_json::Value>,
     /// An optional weight for this spread. Higher weights are given more precedence
     #[serde(skip_serializing_if = "Option::is_none")]
     pub weight: Option<usize>,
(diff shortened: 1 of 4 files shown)


thomastaylor312 (changes requested):
So I think it would probably be best not to do this change. We specifically call out that this scaler matches on host labels that are all strings and this could lead to some more confusing behavior both in code and from what happens for a user. Preferably this would be caught during (our currently non-existent) validation

PR #140, opened 2023-07-28
fix(*): acquire permit for work after receive msg

brooksmtownsend:
Signed-off-by: Brooks Townsend <brooks@cosmonic.com>

## Feature or Problem
The problem with the current implementation is that consumers acquire permits _before_ knowing that work needs to be done. This PR is a simple refactor that ensures that all consumers are awaiting their next message on the stream, and then once they receive the next message they attempt to acquire the permit.

This PR also updates the control interface to 0.28.1, which includes a small bug fix for querying claims when a version field may be null.

## Related Issues
N/A

## Release Information
v0.5.0-alpha.2

## Consumer Impact
This simply makes wadm operate as intended.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
I made sure this works with the default values with more than 256 consumers.

### DIFF
--- a/bin/main.rs
+++ b/bin/main.rs
@@ -72,13 +72,8 @@ struct Args {
 
     /// (Advanced) Tweak the maximum number of jobs to run for handling events and commands. Be
     /// careful how you use this as it can affect performance
-    #[arg(
-        short = 'j',
-        long = "max-jobs",
-        default_value = "256",
-        env = "WADM_MAX_JOBS"
-    )]
-    max_jobs: usize,
+    #[arg(short = 'j', long = "max-jobs", env = "WADM_MAX_JOBS")]
+    max_jobs: Option<usize>,
 
     /// The URL of the nats server you want to connect to
     #[arg(
@@ -193,6 +188,8 @@ async fn main() -> anyhow::Result<()> {
 
     let manifest_storage = nats::ensure_kv_bucket(&context, args.manifest_bucket, 1).await?;
 
+    debug!("Ensuring event stream");
+
     let event_stream = nats::ensure_stream(
         &context,
         EVENT_STREAM_NAME.to_owned(),
@@ -203,6 +200,8 @@ async fn main() -> anyhow::Result<()> {
         ),
     )
     .await?;
+
+    debug!("Ensuring command stream");
 
     let command_stream = nats::ensure_stream(
         &context,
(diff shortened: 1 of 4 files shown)


src/consumers/manager.rs
```diff
@@ -267,12 +267,13 @@ where
     C: Stream<Item = Result<ScopedMessage<W::Message>, async_nats::Error>> + Unpin,
 {
     loop {
+        // Get next value from stream, returning error if the consumer stopped
+        let res = consumer.next().await.ok_or(WorkError::ConsumerStopped)?;
```
thomastaylor312, line 271: This unfortunately could also cause issues. Let's say we've reached the limits on permits, now we've pulled work from the queue and could end up waiting until the timeout for the message (or close to it) and then it will expire. So we'll start doing work and then fail when we try to ack (leading to possible duplication). We need to make sure we can a) get a permit and b) that there is work to pull

thomastaylor312 (changes requested):

brooksmtownsend:
Pushed up one more commit, which comes from @thomastaylor312 and I discussing a bit on a call. Essentially, this solution is complicated to really do correctly, which just comes down to this:
1. Wait for a message to come in that needs work
2. Pull message off the stream _only_ if a permit is available, otherwise wait for a permit
3. Do work

Doing this without polling the future on the stream and thus beginning the jetstream consumer ack timer is difficult, and while we discussed a few possible ways to work around this it would introduce more trouble than it's worth at this time. So, I'm proposing that we make the `max_jobs` argument optional and set the max number to the `Semaphore::MAX_JOBS` limit, effectively making it infinite. Then, we can work on optimizing that number as we test in high-throughput scenarios.

bin/main.rs
```diff
@@ -72,13 +72,8 @@ struct Args {
 
     /// (Advanced) Tweak the maximum number of jobs to run for handling events and commands. Be
     /// careful how you use this as it can affect performance
```
connorsmith256, line 74: Consider filing an issue and noting it here about setting and then exceeding the max job count causes wadm to not process more work

connorsmith256 (approved):

bin/main.rs
brooksmtownsend: https://github.com/wasmCloud/wadm/issues/141 :)

thomastaylor312 (approved):

PR #144, opened 2023-08-03
feat(status): implement status for applications

### LINKED ISSUE #142: [FEAT] Support application status
Applications, after being deployed, currently always just show as the default `Undeployed` status. This was intentional as part of the MVP, but now it's time to implement those.

Application statuses adhere to the following enum.
```rust
/// All possible status types
#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, Clone, Copy, Default)]
#[serde(rename_all = "lowercase")]
pub enum StatusType {
    #[default]
    Undeployed,
    Compensating,
    Ready,
    Failed,
}
```

Undeployed is when an application is put but not deployed, Compensating should be used when actions are being taken to get to desired state, Ready should be used when an application is fully deployed, and Failed should be used when an application fails to deploy.

This feature can be at-minimum implemented by storing the status of a manifest in a store, but it would be best implemented using a NATS stream for statuses so that any individual consumer could watch for status changes and act accordingly.

brooksmtownsend:
## Feature or Problem
This PR implements status for applications by adding a `status` function on the `Scaler` trait, and modifying the status accordingly. The `Undeployed` status is applied by default when a new model is put, or when the model is undeployed after deployment.

The way we calculate status is effectively a little state machine
1. When manifest deployed, set to `Compensating` as initial commands are sent out
2. After the initial set of expected events is received, ask the scaler to reconcile. 
3. If reconciliation returns commands, stay in `Compensating` state and publish those commands, and if the commands list is empty continue.
4. When commands are empty from a scaler, request the status from that scaler. The actor and provider spreadscalers are considered `Ready` if all of the spreads are satisfied, and `Failed` otherwise.
5. Publish the status to the `wadm_status` stream for consumption.
6. Whenever an event is handled by an individual scaler, set status to `Compensating` if commands are sent as a result of that event, or repeat 4.

The `wadm_status` stream will store the last 10 messages for status and those 10 messages will live forever.

## Related Issues
Fixes #142 

## Release Information
v0.5.0

## Consumer Impact
Consumers will see the status actually update from `Undeployed` when querying the wadm API.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
Modified e2e tests to include status assertions, which is the closest real-world example that we have. 

### Manual Verification
I deployed a few manifests using this PR and made sure that some basic examples will result in a failed status, e.g. deploying 2 HTTP servers with only one host:
```
nats req "wadm.api.default.model.list" '{}'
10:24:44 Sending request on "wadm.api.default.model.list"
10:24:44 Received with rtt 6.0045ms
[{"name":"echo","version":"v0.0.2","description":"wasmCloud Echo Example","deployed_version":"v0.0.2","status":"failed","status_message":"Could not satisfy spread default for wasmcloud.azurecr.io/httpserver:0.17.0, 1/2 eligible hosts found."}]
```

### DIFF
--- a/bin/main.rs
+++ b/bin/main.rs
@@ -2,10 +2,7 @@ use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 
-use async_nats::jetstream::{
-    stream::{RetentionPolicy, Stream},
-    Context,
-};
+use async_nats::jetstream::{stream::Stream, Context};
 use clap::Parser;
 use tokio::sync::Semaphore;
 use tracing::log::debug;
@@ -202,9 +199,6 @@ async fn main() -> anyhow::Result<()> {
             "A stream that stores all events coming in on the wasmbus.evt topics in a cluster"
                 .to_string(),
         ),
-        RetentionPolicy::WorkQueue,
-        None,
-        None,
     )
     .await?;
 
@@ -215,20 +209,13 @@ async fn main() -> anyhow::Result<()> {
         COMMAND_STREAM_NAME.to_owned(),
         vec![DEFAULT_COMMANDS_TOPIC.to_owned()],
         Some("A stream that stores all commands for wadm".to_string()),
-        RetentionPolicy::WorkQueue,
-        None,
-        None,
     )
     .await?;
 
-    let status_stream = nats::ensure_stream(
+    let status_stream = nats::ensure_status_stream(
         &context,
         STATUS_STREAM_NAME.to_owned(),
         vec![DEFAULT_STATUS_TOPIC.to_owned()],
-        Some("A stream that stores all status updates for wadm applications".to_string()),
-        RetentionPolicy::Limits,
-        Some(std::time::Duration::from_nanos(0)),
-        Some(10),
     )
     .await?;
 
@@ -249,9 +236,6 @@ async fn main() -> anyhow::Result<()> {
         mirror_stream.to_owned(),
         event_stream_topics.clone(),
         Some("A stream that publishes all events to the same stream".to_string()),
-        RetentionPolicy::WorkQueue,
-        None,
-        None,
     )
     .await?;
 
--- a/bin/main.rs
+++ b/bin/main.rs
@@ -2,7 +2,10 @@ use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 
-use async_nats::jetstream::{stream::Stream, Context};
+use async_nats::jetstream::{
+    stream::{RetentionPolicy, Stream},
+    Context,
+};
 use clap::Parser;
 use tokio::sync::Semaphore;
 use tracing::log::debug;
@@ -17,9 +20,9 @@ use wadm::{
     scaler::manager::{ScalerManager, WADM_NOTIFY_PREFIX},
     server::{ManifestNotifier, Server, DEFAULT_WADM_TOPIC_PREFIX},
     storage::{nats_kv::NatsKvStore, reaper::Reaper},
-    workers::{CommandPublisher, CommandWorker, EventWorker},
+    workers::{CommandPublisher, CommandWorker, EventWorker, StatusPublisher},
     DEFAULT_COMMANDS_TOPIC, DEFAULT_EVENTS_TOPIC, DEFAULT_MULTITENANT_EVENTS_TOPIC,
-    DEFAULT_WADM_EVENTS_TOPIC,
+    DEFAULT_STATUS_TOPIC, DEFAULT_WADM_EVENTS_TOPIC,
 };
 
 mod connections;
@@ -31,6 +34,7 @@ use connections::{ControlClientConfig, ControlClientConstructor};
 
 const EVENT_STREAM_NAME: &str = "wadm_events";
 const COMMAND_STREAM_NAME: &str = "wadm_commands";
+const STATUS_STREAM_NAME: &str = "wadm_status";
 const MIRROR_STREAM_NAME: &str = "wadm_mirror";
 const MULTITENANT_MIRROR_STREAM_NAME: &str = "wadm_multitenant_mirror";
 const NOTIFY_STREAM_NAME: &str = "wadm_notify";
@@ -198,6 +202,9 @@ async fn main() -> anyhow::Result<()> {
             "A stream that stores all events coming in on the wasmbus.evt topics in a cluster"
                 .to_string(),
         ),
+        RetentionPolicy::WorkQueue,
+        None,
+        None,
     )
     .await?;
 
@@ -208,6 +215,20 @@ async fn main() -> anyhow::Result<()> {
         COMMAND_STREAM_NAME.to_owned(),
         vec![DEFAULT_COMMANDS_TOPIC.to_owned()],
         Some("A stream that stores all commands for wadm".to_string()),
+        RetentionPolicy::WorkQueue,
+        None,
+        None,
+    )
+    .await?;
+
+    let status_stream = nats::ensure_stream(
+        &context,
+        STATUS_STREAM_NAME.to_owned(),
+        vec![DEFAULT_STATUS_TOPIC.to_owned()],
+        Some("A stream that stores all status updates for wadm applications".to_string()),
+        RetentionPolicy::Limits,
+        Some(std::time::Duration::from_nanos(0)),
+        Some(10),
     )
     .await?;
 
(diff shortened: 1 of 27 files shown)


bin/main.rs
```diff
@@ -209,6 +216,20 @@ async fn main() -> anyhow::Result<()> {
...
+    .await?;
+
+    let status_stream = nats::ensure_stream(
+        &context,
+        STATUS_STREAM_NAME.to_owned(),
+        vec![DEFAULT_STATUS_TOPIC.to_owned()],
+        Some("A stream that stores all status updates for wadm applications".to_string()),
+        RetentionPolicy::Limits,
+        Some(std::time::Duration::from_nanos(0)),
+        Some(10),
```
brooksmtownsend, line 232: Choosing here to store status forever time-wise, but only retain the last 10 statuses for manifests.

src/scaler/spreadscaler/mod.rs
```diff
@@ -211,15 +219,29 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ActorSpreadScaler<S> {
                     }
                 } else {
                     // No hosts were eligible, so we can't attempt to add or remove actors
-                    // We will want to return an error to indicate an unsatisfied Scaler here, eventually
-                    trace!("Found no eligible hosts");
+                    trace!(?spread.name, "Found no eligible hosts for spread");
+                    spread_status.push(StatusInfo::failed(&format!("Could not satisfy spread {} for {}, 0/1 eligible hosts found.", spread.name, self.config.actor_reference)));
```
connorsmith256, line 223: 0/1? Or just 0? I.e. is only 1 host ever needed?

src/scaler/spreadscaler/provider.rs
```diff
@@ -214,14 +222,38 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ProviderSpreadScaler<S> {
                                 })
                             })
                             .take(num_to_start)
-                            .collect::<Vec<Command>>()
+                            .collect::<Vec<Command>>();
+
+                        if commands.len() < num_to_start {
+                            let msg = format!("Could not satisfy spread {} for {}, {}/{} eligible hosts found.", spread.name, self.config.provider_reference, running_for_spread.len(), count);
```
connorsmith256, line 228: Curious why we compare `commands.len()` to `num_to_start`, but then report `running_for_spread.len()` out of `count`

src/server/handlers.rs
```diff
@@ -75,7 +78,18 @@ impl<P: Publisher> Handler<P> {
             }
         }
 
-        let manifest_name = manifest.metadata.name.clone();
+        let manifest_name = manifest.metadata.name.trim().to_string();
+        if manifest_name.contains(' ') || manifest_name.contains('.') {
```
connorsmith256, line 82: How about a simple regex for this?

src/workers/event.rs
```diff
@@ -709,8 +713,22 @@ where
...
+        } else {
+            StatusInfo::compensating("Model deployed, running initial compensating commands")
+        };
+
+        if let Err(e) = self
+            .status_publisher
+            .publish_status(data.manifest.metadata.name.as_ref(), status)
+            .await
+        {
+            warn!(error = ?e, "Failed to set status to compensating");
```
connorsmith256, line 727: Nit: not always "compensating"

src/workers/event.rs
```diff
@@ -741,17 +769,44 @@ where
...
+
+        // Resolve futures, computing commands for scalers, publishing statuses, and combining any errors
+        let (commands, res) = futures::future::join_all(futs).await.into_iter().fold(
+            (vec![], Ok(())),
+            |(mut cmds, res), (mut new_cmds, new_res)| {
+                cmds.append(&mut new_cmds);
+                let res = match (res, new_res) {
+                    (Ok(_), Ok(_)) => Ok(()),
+                    (Ok(_), Err(e)) | (Err(e), Ok(_)) => Err(e),
+                    (Err(e), Err(e2)) => Err(e.context(e2)),
```
connorsmith256, line 801: Not sure `context` is right here. If there are two unrelated errors won't that result in `err1: err2`? Do you want to join instead?

connorsmith256 (commented):
Looks awesome. A few questions

src/scaler/spreadscaler/mod.rs
brooksmtownsend: For the actor spreadscaler, a single spread will only ever match with one host since it just picks the first host out of the eligible list (see comment just a bit above)
```
        // NOTE(brooksmtownsend) it's easier to assign one host per list of requirements than
        // balance within those requirements. Users should be specific with their requirements
        // as wadm is not responsible for ambiguity, a future scaler like a DaemonScaler could handle this
```

src/scaler/spreadscaler/provider.rs
brooksmtownsend: Yeah, I ran right into this since I reported what we compare at first. Turns out if I wanted to start 2 replicas of a provider, and one is already running, then `num_to_start` is 1 and `commands.len()` is 0. What we really want to report here though is that we already found an eligible host but something is running there, showing `1/2 eligible hosts found`, rather than `0/1`.

src/server/handlers.rs
brooksmtownsend: aw man don't make me regex

brooksmtownsend: In all seriousness, I'm fine doing that as long as whatever was going on with big regular expressions in our other Rust project isn't something that _normally_ happens.

src/workers/event.rs
brooksmtownsend: Is join a separate anyhow Error function? You're right they are two unrelated errors, though I haven't seen something error here yet so it's hard to say how unrelated they would be.

bin/main.rs
thomastaylor312: Yep, this seems smart enough for now

src/scaler/spreadscaler/provider.rs
thomastaylor312: Might want to add a comment here explaining the logic behind it

src/server/handlers.rs
```diff
@@ -761,4 +782,20 @@ impl<P: Publisher> Handler<P> {
...
         self.send_reply(reply, response).await;
     }
+
+    async fn get_manifest_status(&self, lattice_id: &str, name: &str) -> Option<StatusInfo> {
+        match self
+            .status_stream
+            .get_last_raw_message_by_subject(&format!("wadm.status.{lattice_id}.{name}",))
+            .await
+            .map(|raw| {
+                B64decoder
```
thomastaylor312, line 792: Question for my edification: Why do we have to b64 decode the body here?

src/workers/event_helpers.rs
```diff
@@ -95,6 +95,37 @@ impl LinkSource for wasmcloud_control_interface::Client {
...
+pub struct StatusPublisher<Pub> {
+    publisher: Pub,
+    // Topic prefix, e.g. wadm.status.default
+    topic: String,
+}
+
+impl<Pub> StatusPublisher<Pub> {
+    /// Creates an new status publisher configured with the given publisher that will send to the
+    /// specified topic
+    pub fn new(publisher: Pub, topic: &str) -> StatusPublisher<Pub> {
```
thomastaylor312, line 109: Nit: Might want to name the parameter (and the data member of the struct) to be `topic_prefix` so it is just a bit more clear what it is used for

src/server/handlers.rs
connorsmith256: I _think_ that was because said regex was a bit more monstrous. How about this? `[-\w]+`

src/workers/event.rs
connorsmith256: No, I don't think a helper exists for this :/ I was thinking something like `=> anyhow!("{e}, {e2}")`

src/server/handlers.rs
brooksmtownsend: This is because of the `get_last_raw_message` method giving us back b64 encoded values, but if I can set `direct_get` on the stream to `true` I won't have to do this 🤔 Will investigate

PR #146, opened 2023-08-04
feat: manifest and application upgrades

### LINKED ISSUE #143: [FEAT] Support version upgrades for actors and providers
Currently, the Spreadscaler only validates that a specific public key (and link name, for providers) is running with the specified number of replicas on a host. Version upgrades would add more logic into this validation, requiring that a specific version of a resource is deployed as well. This version can be sourced from the OCI reference and doesn't need to check the claims specifically.

This will be a little bit complicated due to the relationships we have with OCI versions and running duplicate providers. TL;DR:
1. Upgrading an actor will require wadm to stop all previously existing instances of an actor, and then start new instances. There is an update actor command in the control interface but it requires that revisions always be greater than the previous revision, which may or may not work well.
2. Upgrading a provider requires wadm to stop the previously existing instance of the provider and starting a new one.

Scalers don't have a mechanism to say what order to send out commands in, so this might also be subject to race conditions with upgrades. It would help if the host has a robust mechanism for upgrading actors, providers, and maybe link definitions, but that requires work there. 

I need to do a little more research to see if link definitions can upgrade too.

### LINKED ISSUE #145: [BUG] Newly deployed applications do not terminate old resources
When you create a new version of an application, we create entirely new Spreadscaler instances that handle computing state and launching new resources according to the latest version. However, old resources that are no longer present aren't removed when upgrading the application. Consider the following manifest for KV counter:

```yaml
kind: Application
metadata:
  name: kvcounter
  annotations:
    version: v0.0.1
    description: "wasmCloud Key Value Counter Example"
spec:
  components:
    - name: kvcounter
      type: actor
      properties:
        image: wasmcloud.azurecr.io/kvcounter:0.4.2
      traits:
        - type: spreadscaler
          properties:
            replicas: 1
        - type: linkdef
          properties:
            target: redis
            values:
              URL: redis://127.0.0.1:6379/
        - type: linkdef
          properties:
            target: httpserver
            values:
              ADDRESS: 0.0.0.0:8081

    - name: httpserver
      type: capability
      properties:
    - name: redis
      type: capability
      properties:
        image: wasmcloud.azurecr.io/kvredis:0.21.0
        contract: wasmcloud:keyvalue
```

Now, if I were to swap out my capability of redis for kv-vault for example,
```
kind: Application
metadata:
  name: kvcounter
  annotations:
    version: v0.0.1
    description: "wasmCloud Key Value Counter Example"
spec:
  components:
    - name: kvcounter
      type: actor
      properties:
        image: wasmcloud.azurecr.io/kvcounter:0.4.2
      traits:
        - type: spreadscaler
          properties:
            replicas: 1
        - type: linkdef
          properties:
            target: vault
        - type: linkdef
          properties:
            target: httpserver
            values:
              ADDRESS: 0.0.0.0:8081

    - name: httpserver
      type: capability
      properties:
    - name: vault
      type: capability
      properties:
        image: wasmcloud.azurecr.io/kv-vault:0.4.1
        contract: wasmcloud:keyvalue
```

What I would expect to happen is
1. New manifest is deployed
2. Redis provider isn't present in manifest, so it's removed
3. Link definition to redis is deleted
4. KV vault provider is started
5. KV vault provider health check passes, link is put between kvcounter and vault

Steps 2 and 3 above are the new things that must be added to address this issue. I imagine the implementation of this will include adding some logic to compare the old application scalers to the new ones, and anything that's present in old that's not present in the new should be deleted. Note that scalers _can_ handle changing the amount of replicas of an actor or provider, it's just when you modify to use a different actor or provider (or version #143) that the issue comes into play.

brooksmtownsend:
## Feature or Problem
This PR changes the manifest deployment logic to look at old scaler components, compare them to new components, and cleanup the resources if the components no longer exist. Refer to the example in #145 for when this is appropriate.

This PR also (in the second commit) implements version upgrades by considering Actor and Provider spreadscalers meaningfully different if their image reference is different. Doing this upgrade properly _does_ seem to depend on a short delay between cleaning up old resources and starting new ones, and before this PR is accepted I'd like to resolve that instead of sleeping. Ideally, I would like to proceed with reconciliation immediately and _only_ reconcile commands for scalers that are completely new, allowing scalers to instead launch new resources based off of the events from the old resources stopping. I'm 80% confident that this will work as intended, and I'm pretty sure the only thing preventing this from working at the time of writing this is the fact that scalers don't know those events are for the resources that are stopping 🤔 

NOTE: Link definition upgrades work _only_ if the actor or provider reference changes because of the ID structure. I don't want to encode each and every value into the linkdef scaler ID, but maybe I'll have to 🤔 

## Related Issues
Fixes #145 
Fixes #143 

## Release Information
v0.5.0 / v0.5.0-alpha.4

## Consumer Impact
If consumers expected old resources to stick around with new manifest versions, they will now be cleaned up.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
I manually verified this works

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        e2e_test: [e2e_multiple_hosts, e2e_multitenant]
+        e2e_test: [e2e_multiple_hosts, e2e_multitenant, e2e_upgrades]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -9,6 +9,10 @@ jobs:
   test:
     name: e2e
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        e2e_test: [e2e_multiple_hosts, e2e_multitenant]
 
     steps:
       - uses: actions/checkout@v2
@@ -25,17 +29,19 @@ jobs:
         with:
           key: "ubuntu-22.04-rust-cache"
 
-      # Run all tests
-      - name: Run tests
+      # Run e2e tests in a matrix for efficiency
+      - name: Run tests ${{ matrix.e2e_test }}
         id: test
-        run: make test-e2e
-    
+        env:
+          WADM_E2E_TEST: ${{ matrix.e2e_test }}
+        run: make test-individual-e2e
+
       # if the previous step fails, upload logs
       - name: Upload logs for debugging
         uses: actions/upload-artifact@v2
         if: ${{ failure() && steps.test.outcome == 'failure' }}
         with:
-          name: e2e-logs
+          name: e2e-logs-${{ matrix.e2e_test }}
           path: ./test/e2e_log/*
           # Be nice and only retain the logs for 7 days
           retention-days: 7
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,7 @@ ifeq ($(shell nc -czt -w1 127.0.0.1 4222 || echo fail),fail)
 	@$(MAKE) build
 	RUST_BACKTRACE=1 $(CARGO) test --test e2e_multitenant --features _e2e_tests --  --nocapture 
 	RUST_BACKTRACE=1 $(CARGO) test --test e2e_multiple_hosts --features _e2e_tests --  --nocapture 
+	RUST_BACKTRACE=1 $(CARGO) test --test e2e_upgrades --features _e2e_tests --  --nocapture 
 else
 	@echo "WARN: Not running e2e tests. NATS must not be currently running"
 	exit 1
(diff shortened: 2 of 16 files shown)


brooksmtownsend:
Should be merged after #146

brooksmtownsend:
This PR history looks _wild_ but it's just because I was continuously rebasing off of #146

brooksmtownsend:
As a note, I do notice some flakiness in the e2e tests around status updates in this PR. I'd like to tackle status assertion flakiness in a future PR.

connorsmith256 (approved):
Love the new tests

PR #150, opened 2023-08-10
add certs back

jordan-rash:
## Feature or Problem
Need to run CA Certs update for wadm to work with NGS

I dont have docker on this machine, so im letting CI sanity check it

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM debian:bullseye-slim AS base
 
-RUN update-ca-certificates
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y ca-certificates
 
 FROM base AS base-amd64
 ARG BIN_AMD64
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM debian:bullseye-slim AS base
 
+RUN update-ca-certificates
+
```
brooksmtownsend, line 4: Will this work if ca-certificates aren't installed? And don't we have to do this on the final container (e.g. after line 13)?

brooksmtownsend: Sadly, no Docker in CI I don't think 😬

jordan-rash: The first container is the base container, so running it in there should trickle into the others.....I need to go find another laptop to check it...ill report back

jordan-rash:
This should work brooks

brooksmtownsend (approved):

Issue #149, opened 2023-08-10
[BUG] E2E tests are flaky when asserting status


brooksmtownsend:
You can take a look at a few of the latest action runs and see the e2e multitenant / multiple host tests fail occasionally on the status assertions. This may just be a mismatch in the e2e tests where we need to be more careful when asserting things, or it might be edge cases that the e2e tests are catching around undeploying an app and then a new event ends up changing the status unintentionally.

I am having trouble reproducing the issues locally but that's just CI life 😄

PR #152, opened 2023-08-10
feat(status): Fully compute status when publishing status updates

brooksmtownsend:
Fixes https://github.com/wasmCloud/wadm/issues/149

This PR is a simplification of the status algorithm which should make manifest statuses more consistently correct. Essentially, when an event gets handled by a set of scalers, it follows this flow:
image

This is a little less efficient in terms of calculating the exact status of an individual scaler and publishing updates based on that, but it's simple for now and we can improve the algorithm as we see a good opportunity to do so. If you only ever ran one wadm instance this PR wouldn't be necessary, but as you can see from our e2e tests it results in a lot of flakiness when running multiple instances. Storing an internal property of `status` in a scaler works nice as long as that scaler is handling all of the relevant events, which when running multiple instances that is not guaranteed.

I am unable to make e2e tests fail locally because of statuses using this code. Very occasionally I believe an issue with storing the heartbeat will cause a leftover link definition, actor, or provider, but that's not related to this PR.

### DIFF
--- a/src/scaler/spreadscaler/mod.rs
+++ b/src/scaler/spreadscaler/mod.rs
@@ -58,6 +58,7 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ActorSpreadScaler<S> {
     }
 
     async fn status(&self) -> StatusInfo {
+        let _ = self.reconcile().await;
         self.status.read().await.to_owned()
     }
 
--- a/src/workers/event.rs
+++ b/src/workers/event.rs
@@ -742,14 +742,13 @@ where
         )
         .await;
 
-        trace!(?commands, "Publishing commands");
-
         let status = if commands.is_empty() {
             scaler_status(&scalers).await
         } else {
             StatusInfo::compensating("Model deployed, running initial compensating commands")
         };
 
+        trace!(?status, "Setting status");
         if let Err(e) = self
             .status_publisher
             .publish_status(data.manifest.metadata.name.as_ref(), status)
@@ -759,6 +758,7 @@ where
         };
 
         // Now handle the result from reconciliation
+        trace!(?commands, "Publishing commands");
         self.command_publisher.publish_commands(commands).await?;
 
         res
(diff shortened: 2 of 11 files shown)


src/workers/event.rs
```diff
@@ -944,9 +950,12 @@ where
...
                             warn!(error = ?e, "Failed to set status to undeployed");
                         }
-                        Ok(None)
+                        return message.ack().await.map_err(WorkError::from);
+                    }
+                    Some(Err(e)) => {
+                        message.nack().await;
+                        return Err(WorkError::Other(e.into()));
                     }
-                    Some(Err(e)) => Err(e),
```
brooksmtownsend, line 949: This is an optimization, no scalers need to handle the ManifestUnpublished message so we can just ack / nack here instead of running _every_ scaler.

src/scaler/spreadscaler/mod.rs
```diff
@@ -58,6 +58,7 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ActorSpreadScaler<S> {
     }
 
     async fn status(&self) -> StatusInfo {
+        let _ = self.reconcile().await;
```
thomastaylor312, line 61: My concern with doing it this way is that a reconcile just to get the status could theoretically cause a backoff period to occur:

1. Reconcile triggers
2. An event we are waiting for doesn't arrive or is missed
3. We wait the full 30s for actor scaler to backoff

I don't think this is blocking but I have an unproven worry this could cause some cases where wadm would appear "slow" to reconcile

thomastaylor312 (approved):

src/scaler/spreadscaler/mod.rs
brooksmtownsend: I think your specific concern is alleviated since this is inside of the Spreadscaler itself, we are not causing a backoff because it's not a Backoff aware spreadscaler. I do agree with you though that we have to be pretty careful about reconciling the generic Scaler trait.

Issue #123, opened 2023-06-02
Don't fight over linkdef configuration


thomastaylor312:
Right now if you deploy a manifest that has a linkdef between the same actor and provider as another manifest, and the values for that linkdef are different, the scalers will end up in an infinite loop fight as they notice that the values they are expecting are different. I have mitigated this in an upcoming PR, but we should probably have a way to indicate that a linkdef is managed by wadm. If it is managed by wadm, then the scaler will not try to delete and recreate if the values are different, but if it isn't managed, wadm can "take control" and set the right values

autodidaddict:
For the time being, before we figure out an alternative to the way link definitions work, we should probably reject manifests that could create a conflicting link definition. @thomastaylor312 @connorsmith256 we talked about this a little bit earlier today. 

There are some long term design concerns that I think might make this situation less of a problem, but for now we should prevent users from creating an unstable and unpredictable environment.

I'd like to propose that wadm reject any manifest that contains a _different_ link definition for a previous link between the same actor, provider ID, and link name. If two manifests supply the same link definition with the same values, then wadm can just reuse it.

So, by way of example:
* Application A
    * Link from KeyValueCounter to KvRedis/default with value URL=localhost:6379
* Application B
    * Link from KeyValuecounter to KvRedis/default with value URL=localhost:6379
* Application C
    * Link from KeyValueCounter to KvRedis/default with value URL=superexample:6378  

If these application manifests were added in sequence, then Application A would deploy just fine. Application B would deploy and "reuse" (no-op) the previously existing link because the values are _identical_. Application C would be rejected because it attempts to define a conflicting link definition.

This does end up in a "first wins" scenario, where the order in which the manifests are submitted dictates which ones might be rejected, but at least it still prevents the scenario we want to avoid.

lee-orr:
@autodidaddict @thomastaylor312 @brooksmtownsend @connorsmith256 Can I ask for more information about what the long term solutions for this would be? I've been trying to get wasmCloud to work on a home server, with the plan of hosting multiple apps on a single lattice - some of which would be using identical sets of elements (actor images + providers) with different values on the links. This would be, for example, to allow me to have multiple different static file hosts that provide different files based on port (or ideally, host name for the URL - if I have the time to contribute that to the http server), or for utilizing a single auth actor image to provide auth for different services at different endpoints.

thomastaylor312:
[...]

Issue #122, opened 2023-06-02
Validate manifests before accepting them


thomastaylor312:
Right now we blindly accept manifests without any validation. This causes issues where you might have misplaced something like your spread definitions and also leads to manifests fighting with each other (if two different manifests link to the same provider with different values, so those scalers each try to compete to set their version of the linkdef). At the minimum, validation should probably include:

- Collision with other manifests, particularly with actor/provider versions, and linkdefs with different values for the same actor
- Unrecognized or missing fields when parsing the manifest

thomastaylor312:
One possible solution here is to define a json schema and use that: https://docs.rs/jsonschema/latest/jsonschema/

brooksmtownsend:
Consider #123 and @autodidaddict's scenarios when validating the manifests

brooksmtownsend:
One more comment, I think this issue should cover the problem outlined in https://github.com/wasmCloud/wadm/issues/131 and reject non-string label requirements.

aish-where-ya:
Here are some of the cases covered in #153 

- Multiple actors or providers cannot have the same names in a single manifest
- A linkedef - provider reference pair needs to be unique in a capability provider definition.
- Linkdefs in a manifest should have a unique target

The next PR will cover validation using a JSONSchema for manifest items.

PR #153, opened 2023-08-11
Added validation support for manifests [WIP]

aish-where-ya:
## Feature or Problem
Manifest validation:

- Validate manifests against a JSONSchema (PENDING creation of a JSON schema against which validation can be performed)
- Actors and providers cannot have the same names in a single manifest
- Actors or providers with different names cannot have the same image reference.
- Linkdefs in a manifest should have a unique target
- Linkdefs in the current manifest having the same origin and target as a previous version of the manifest will be rejected if they have different values provided to the new linkdef.
- Tests for these cases (PENDING WIP)

## Related Issues
#122 #123 

## Release Information
next

## Consumer Impact
Invalid manifests will be rejected

## Testing
NA

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
[WIP]

### DIFF
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -61,33 +61,6 @@ impl Manifest {
             .get(DESCRIPTION_ANNOTATION_KEY)
             .map(|v| v.as_str())
     }
-
-    /// Return a list of the linkdefs in the manifests
-    pub fn get_linkdef_map(&self) -> HashMap<(String, String), HashMap<String, String>> {
-        // Map of the type Component Name -> Target : linkdef values
-        let mut linkdef_map: HashMap<(String, String), HashMap<String, String>> = HashMap::new();
-        for component in &self.spec.components {
-            if let Some(traits_vec) = &component.traits {
-                for trait_item in traits_vec.iter() {
-                    if let Trait {
-                        properties:
-                            TraitProperty::Linkdef(LinkdefProperty {
-                                target: target_name,
-                                values: Some(linkdef_values),
-                            }),
-                        ..
-                    } = &trait_item
-                    {
-                        linkdef_map.insert(
-                            (component.name.to_string(), target_name.to_string()),
-                            linkdef_values.clone(),
-                        );
-                    }
-                }
-            }
-        }
-        linkdef_map
-    }
 }
 
 /// The metadata describing the manifest
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -61,6 +61,33 @@ impl Manifest {
             .get(DESCRIPTION_ANNOTATION_KEY)
             .map(|v| v.as_str())
     }
+
+    /// Return a list of the linkdefs in the manifests
+    pub fn get_linkdef_map(&self) -> HashMap<(String, String), HashMap<String, String>> {
+        // Map of the type Component Name -> Target : linkdef values
+        let mut linkdef_map: HashMap<(String, String), HashMap<String, String>> = HashMap::new();
+        for component in &self.spec.components {
+            if let Some(traits_vec) = &component.traits {
+                for trait_item in traits_vec.iter() {
+                    if let Trait {
+                        properties:
+                            TraitProperty::Linkdef(LinkdefProperty {
+                                target: target_name,
+                                values: Some(linkdef_values),
+                            }),
+                        ..
+                    } = &trait_item
+                    {
+                        linkdef_map.insert(
+                            (component.name.to_string(), target_name.to_string()),
+                            linkdef_values.clone(),
+                        );
+                    }
+                }
+            }
+        }
+        linkdef_map
+    }
 }
 
 /// The metadata describing the manifest
--- a/src/server/handlers.rs
+++ b/src/server/handlers.rs
@@ -1,11 +1,15 @@
 use async_nats::{Client, Message};
+use jsonschema::{Draft, JSONSchema};
 use serde_json::json;
+use std::collections::HashSet;
+use std::fs::File;
+use std::io::BufReader;
 use tracing::{debug, error, instrument, trace};
 
 use crate::{
     model::{
-        internal::StoredManifest, CapabilityConfig, CapabilityProperties, Properties,
-        LATEST_VERSION,
+        internal::StoredManifest, ActorProperties, CapabilityConfig, CapabilityProperties,
+        LinkdefProperty, Manifest, Properties, Trait, TraitProperty, LATEST_VERSION,
     },
     publisher::Publisher,
 };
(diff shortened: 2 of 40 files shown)


src/server/handlers.rs
```diff
@@ -762,3 +747,158 @@ impl<P: Publisher> Handler<P> {
         self.send_reply(reply, response).await;
     }
 }
+
+pub async fn open_schema_file() -> anyhow::Result<BufReader<File>> {
+    let schema_file = File::open("./oam/oam.schema.json")?;
```
brooksmtownsend, line 752: I don't see this file committed, did it miss the git commit?

src/server/handlers.rs
```diff
@@ -762,3 +747,158 @@ impl<P: Publisher> Handler<P> {
...
+    Ok(BufReader::new(schema_file))
+}
+
+// Manifest validation
+pub(crate) fn validate_manifest(
+    manifest: Manifest,
+    current_manifests: StoredManifest,
+) -> Option<String> {
```
brooksmtownsend, line 760: nit: I don't find this too bad, but I think it could be more clear here to use a `Result<()>` and return an error if the validation fails. That way, an assertion like `validate_manifest().is_ok()` would mean it's valid, over `validate_manifest().is_none()`

Not hard set on this and if you disagree, a doc comment indicating the purpose of the return type would be enough.

src/server/handlers.rs
```diff
@@ -762,3 +747,158 @@ impl<P: Publisher> Handler<P> {
...
+    // let reader = open_schema_file().await.unwrap();
+    // let json_schema = serde_json::from_reader(reader).unwrap();
+    // let json_instance = serde_json::to_value(manifest.clone()).unwrap();
+    // let compiled_schema = JSONSchema::options()
```
brooksmtownsend, line 764: Looks like we may not have a JSON schema yet, depending on your answer above. If we don't, I would move this to a separate issue and remove the commented code

src/server/handlers.rs
```diff
@@ -762,3 +747,158 @@ impl<P: Publisher> Handler<P> {
...
+            if !component_details.insert(image_name.to_string()) {
+                return Some(format!(
+                    "Duplicate image reference in manifest: {}",
+                    image_name
+                ));
+            }
+
+            if !link_name_set.insert(link.to_string()) {
+                return Some(format!("Duplicate link name in manifest: {}", image_name));
+            }
```
brooksmtownsend, line 822: This check actually needs to happen together for capability providers, e.g. the combination of the image name and link name needs to be unique, but if one of those differs then that's an issue.

Consider the following:
OK:
- image ref: httpserver:0.17.0, link name `default`
- image ref: httpserver:0.17.0, link name `notdefault`

OK:
- image ref: httpserver:0.17.0, link name `default`
- image ref: kvredis:0.17.0, link name `default`

Not OK:
- image ref: httpserver:0.17.0, link name `default`
- image ref: httpserver:0.17.0, link name `default`

src/server/handlers.rs
```diff
@@ -762,3 +747,158 @@ impl<P: Publisher> Handler<P> {
...
+            properties: ActorProperties { image: image_name },
+        } = &component.properties
+        {
+            if !component_details.insert(image_name.to_string()) {
+                return Some(format!(
+                    "Duplicate image reference in manifest: {}",
+                    image_name
+                ));
+            }
+        }
```
brooksmtownsend, line 836: Looks like you also need to ensure the name is unique

src/server/handlers.rs
```diff
@@ -762,3 +747,158 @@ impl<P: Publisher> Handler<P> {
...
+                },
+        } = &component.properties
+        {
+            if let Err(e) = serde_json::to_string(data) {
+                return Some(format!(
+                    "Unable to serialize JSON config data for component {}: {e:?}",
+                    component.name
+                ));
+            }
+
```
brooksmtownsend, line 812: the provider check should also ensure a unique component name

src/server/handlers.rs
```diff
@@ -762,3 +747,158 @@ impl<P: Publisher> Handler<P> {
...
+                            .eq(linkdef_values)
+                            || existing_linkdefs
+                                .get(&(target_name.to_string(), component.name.clone()))
+                                .map(|x| x.to_owned())
+                                .eq(linkdef_values))
+                    {
+                        return Some(format!(
+                            "Component {} and target {} have different linkdef values from earlier version of manifest",
+                            component.name, target_name
+                        ));
```
brooksmtownsend, line 873: This is actually an accepted scenario, but components and targets cannot have different values from _other_ existing deployed manifests 🤔 This is a really complicated check to do at validation time and I'm leaning towards leaving it out, we should pair on this later.

@thomastaylor312 do you have thoughts here? It's hard to check _every_ other manifest in the store in order to ensure link values are unique, especially because component/targets can have different names, image refs, etc

src/server/handlers.rs
```diff
@@ -762,3 +747,158 @@ impl<P: Publisher> Handler<P> {
...
+                    }
+                }
+            }
+        }
+    }
+    None
+}
+
+#[cfg(test)]
+mod test {
```
brooksmtownsend, line 883: I'd love to see some example manifests that fail the validation based on the different conditions above

brooksmtownsend (changes requested):
A couple requests, overall the logic to validate looks really clean and returns clear error messages. Biggest request here I think is to either include the JSON schema if you have one, or file an issue to create one in the future

src/server/handlers.rs
aish-where-ya: No I forgot to commit it. Sorry.

src/server/handlers.rs
aish-where-ya: Are you referring to the first item in the loop?
```
// Component name validation : each component (actors or providers) should have a unique name
if !component_details.insert(component.name.clone()) {
             return Some(format!(
                 "Duplicate component name in manifest: {}",
                 component.name
             ));
         }
```

brooksmtownsend: NVM, the other check handles this!

src/server/handlers.rs
brooksmtownsend: NVM, an earlier check handles this!

src/model/mod.rs
```diff
@@ -61,6 +61,33 @@ impl Manifest {
             .get(DESCRIPTION_ANNOTATION_KEY)
             .map(|v| v.as_str())
     }
+
+    /// Return a list of the linkdefs in the manifests
+    pub fn get_linkdef_map(&self) -> HashMap<(String, String), HashMap<String, String>> {
+        // Map of the type Component Name -> Target : linkdef values
```
brooksmtownsend, line 67: @aish-where-ya we can remove this now that it's unused, right?

brooksmtownsend (changes requested):
Love the test cases and logic here! Just two final requests to remove an unused function and to add a comment

src/server/handlers.rs
brooksmtownsend: Decided after syncing that this is too complicated to check in the scope of this PR, which is a really nice `validate_manifest` function that can work without querying _every_ other manifest.

brooksmtownsend (approved):

PR #157, opened 2023-08-14
Add Actor and Provider Daemonscalers

brooksmtownsend:
## Feature or Problem
This PR adds a new scaler type, the `daemonscaler`, which is a drop-in replacement for the `spreadscaler` trait. Instead of spreading a certain number of replicas across a specified set of constraints, the `daemonscaler` spreads a certain number of replicas across _all_ hosts in a lattice that match the requirements of the spread.

Kubernetes developers can think of a Spreadscaler as a k8s deployment with label selectors, and the Daemonscaler as a k8s DaemonSet (with optional label selectors)

More documentation will be needed here (PR forthcoming), but just so you know:
1. The daemonscaler uses the exact same config as the spreadscaler
2. Both daemonscalers ignore `weight` entirely
3. The provider spreadscaler ignores the `replicas` field, as you can only run one copy of a provider per host. 

## Related Issues
N/A

## Release Information
v0.6.0

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Added unit tests for the actor daemonscaler to ensure spreading logic works

### Manual Verification
I manually tested this working with and without constraints.

### DIFF
--- a/src/scaler/daemonscaler/mod.rs
+++ b/src/scaler/daemonscaler/mod.rs
@@ -1,0 +1,1041 @@
+use std::cmp::Ordering;
+
+use anyhow::Result;
+use async_trait::async_trait;
+use tokio::sync::{OnceCell, RwLock};
+use tracing::{instrument, trace};
+
+use crate::events::HostHeartbeat;
+use crate::model::Spread;
+use crate::scaler::spreadscaler::{eligible_hosts, spreadscaler_annotations};
+use crate::server::StatusInfo;
+use crate::{
+    commands::{Command, StartActor, StopActor},
+    events::{Event, HostStarted, HostStopped},
+    model::{SpreadScalerProperty, TraitProperty},
+    scaler::Scaler,
+    storage::{Actor, Host, ReadStore},
+};
+
+// pub mod provider;
+
+// Annotation constants
+pub const ACTOR_DAEMON_SCALER_TYPE: &str = "actordaemonscaler";
+
+/// Config for an ActorDaemonScaler
+#[derive(Clone, Debug)]
+struct ActorSpreadConfig {
+    /// OCI, Bindle, or File reference for an actor
+    actor_reference: String,
+    /// Lattice ID that this DaemonScaler monitors
+    lattice_id: String,
+    /// The name of the wadm model this DaemonScaler is under
+    model_name: String,
+    /// Configuration for this DaemonScaler
+    spread_config: SpreadScalerProperty,
+}
+
+/// The ActorDaemonScaler ensures that a certain number of replicas are running on every host, according to a
+/// [SpreadScalerProperty](crate::model::SpreadScalerProperty)
+///
+/// If no [Spreads](crate::model::Spread) are specified, this Scaler simply maintains the number of replicas
+/// on every available host.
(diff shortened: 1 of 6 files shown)


brooksmtownsend:
Note to self: check to make sure this actually works when no spreads are defined

src/scaler/daemonscaler/mod.rs
```diff
@@ -0,0 +1,1038 @@
...
+                    Some(
+                        actors_per_host
+                            .iter()
+                            .filter_map(|(host_id, current_count)| {
+                                // Here we'll generate commands for the proper host depending on where they are running
+                                match current_count.cmp(&self.config.spread_config.replicas) {
+                                    Ordering::Equal => None,
+                                    // Start actors to reach desired replicas
+                                    Ordering::Less => {
+                                        // Right now just start on the first available host. We can be smarter about it later
```
thomastaylor312, line 189: Is this comment a copy/paste leftover? We start on all matching hosts so I figure this doesn't apply

src/scaler/daemonscaler/mod.rs
```diff
@@ -0,0 +1,1038 @@
...
+                }
+            })
+            .flatten()
+            .collect::<Vec<Command>>();
+        trace!(?commands, "Calculated commands for actor daemon scaler");
+
+        let status = match (spread_status.is_empty(), commands.is_empty()) {
+            (true, true) => StatusInfo::ready(""),
+            (_, false) => StatusInfo::compensating(""),
+            (false, true) => StatusInfo::failed(
```
thomastaylor312, line 234: Nit: I don't think I realized this before but this is a really odd way to identify an error case. I don't think it needs to be solved now, but we might want to reevaluate how to indicate errors

src/scaler/daemonscaler/provider.rs
```diff
@@ -0,0 +1,259 @@
...
+                        anyhow::anyhow!(
+                            "Couldn't find a provider id for the provider reference {}",
+                            self.config.provider_reference
+                        )
+                    })
+            })
+            .await
+            .map(|id| id.as_str())
+    }
+}
```
thomastaylor312, line 259: Did you want to add some tests here?

thomastaylor312 (commented):
Just one minor question around tests. Others are nits

src/scaler/daemonscaler/mod.rs
brooksmtownsend: Yes, good catch

src/scaler/daemonscaler/provider.rs
brooksmtownsend: Yes 👀

thomastaylor312 (approved):

thomastaylor312 (approved):

thomastaylor312 (approved):

PR #158, opened 2023-08-15
fix(upgrades): run cleanup commands after adding scalers

brooksmtownsend:
## Feature or Problem
This PR addresses a race condition where, during an upgrade, we published cleanup commands _before_ notifying other wadm instances that they need to upgrade their scalers. This would manifest itself by stopping old resources and then immediately restarting those old resources before the new scalers are created, which effectively makes the upgrade fail.

## Related Issues
N/A

## Release Information
v0.5.0-alpha.5

## Consumer Impact
Users of alpha.4 may see some inconsistent upgrades

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/src/workers/event.rs
+++ b/src/workers/event.rs
@@ -703,7 +703,7 @@ where
             .await;
         let scalers = self.scalers.scalers_for_manifest(&data.manifest);
 
-        if let Some(old_scalers) = old_scalers {
+        let cleanup_commands = if let Some(old_scalers) = old_scalers {
             // This relies on the idea that an ID is a unique identifier for a scaler, and any
             // change in the ID is indicative of the fact that the scaler is outdated and should be cleaned up.
             let (_updated_component, outdated_component): (ScalerList, ScalerList) = old_scalers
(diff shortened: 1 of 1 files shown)


connorsmith256 (commented):
I think I need help understanding this change. For example, now that we wait to send the old commands, does `self.scalers.add_scalers` have any effect?

brooksmtownsend:
> I think I need help understanding this change. For example, now that we wait to send the old commands, does `self.scalers.add_scalers` have any effect?

The whole philosophy behind the quick upgrades PR was that the replacement of a manifest (new deploy) would follow this process:
1. Remove the old scalers, preventing them from taking any more compensating actions for events
2. Split the scalers into "updated" components, which contain small changes that a scaler can handle internally (replica change, for example) and "outdated" components where the scaler should cleanup entirely. This will remove outdated resources that no longer exist, AND remove resources that _have_ to update, like when a provider updates its image reference.
3. Add the scalers from the new manifest, which runs its initial reconcile (updating the updated components)
4. Events from cleaning up the "outdated" scalers start to flow in, which the new scalers recognize as relevant and start the new resources. 

With manifests that contain more than just a couple of scalers, there's a race condition here between 2 and 3 where the commands start to take affect before the new scalers are ready, so there is a decent bit of downtime (usually waiting for a host heartbeat) before the new resources would start. Effectively, this PR swaps step 2 and 3 above.

For example:
1. Manifest v1 has HTTPserver v1, and v2 has HTTPserver v2
2. We remove scalers per step 1 above
3. We compute the cleanup commands, which will stop HTTP v1, but we don't publish yet
4. We add the new scalers and compute their relevant commands (the provider scaler sees that _an_ HTTP server is deployed based on public key, so it's satisfied for now)
5. We publish the cleanup commands, which results in a ProviderStopped event
6. The v2 scaler sees the ProviderStopped event and tells the host to start HTTPserver v2
7. Upgrade done

connorsmith256 (approved):

Issue #125, opened 2023-06-05
Figure out how to remove all jitter when running multiple wadm instances


thomastaylor312:
The work in #112 removed a large part of the jitter issue, but it still can occur when running multiple wadm instances due to the delay in a notify event being handled by other wadm instances. We should find a way to make sure that all (or at least most) of the wadm instances all enter waiting/backoff mode before handling events. There are two different ways I can think of right now to handle this:

1. Some sort of CRDT for storing expected events so we know everything is working from the same state
2. Instead of just firing off the notification, we switch to a request/response pattern. So when one instance notifies, it will wait up to something like 300ms to get responses. We won't be able to guarantee that _everything_ gets it, but we will know that at least some of them got it before continuing on. 
    - Doing this with Jetstream could work but it'd take a lot of work because the message has to deliver to everyone and we have to make sure they don't process events before they receive the message. 

I think 1 would be more robust of a solution, but will be much more complex as we'd have to rely on preexisting CRDT stuff (which does exist for Rust, but I haven't tested it). 2 is obviously easier, but could still be racy as we can't guarantee that we've reached everyone else (as wadm doesn't know if there are other wadms running or not)

PR #171, opened 2023-09-08
Use actor scale ctl commands

brooksmtownsend:
## Feature or Problem
This PR refactors the spreadscaler and daemonscaler to use the ScaleActor command instead of computing differences and using StopActor / StartActor. The host itself already computes this when using scale, and this will result in just generally predictable deployments with wadm.

## Related Issues
#125
Annotations test will properly pass after https://github.com/wasmCloud/wasmCloud/pull/656

## Release Information
v0.7.0

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Modified all unit tests for actor scalers to expect proper scale commands. The test failure in the unit test is actually the correct test but will only pass after the issue above is fixed. (Confirmed the above PR _does_ fix the issue here!)


thomastaylor312 (commented):
Initial implementation here looks great! This should really simplify things

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "wadm"
 description = "wasmCloud Application Deployment Manager: A tool for running Wasm applications in wasmCloud"
-version = "0.6.0"
+version = "0.6.1"
```
brooksmtownsend, line 4: Note: We decided to change this to 0.7.0 if I recall correctly @thomastaylor312

thomastaylor312 (dismissed):

Cargo.toml
thomastaylor312: Yes that is correct. Feel free to bump here, but I know I have one PR I should open today

thomastaylor312 (approved):

PR #172, opened 2023-09-11
fix: implement model name sort in model listing

### LINKED ISSUE #156: [FEAT] Sort applications by name for `wash app list`
This is really just a nit, but I notice that repeated calls to `wash app list` can return apps in a different order. It would be nice to just get that in a predictable, alphabetical order.
```
~/tmp ➜ wash app list
  Name        Latest Version   Description                           Deploy Status
  kvcounter   v0.0.1           wasmCloud Key Value Counter Example      Undeployed
  echo        v0.0.1           wasmCloud Echo Example                 Compensating

~/tmp ➜ wash app list


  Name        Latest Version   Description                           Deploy Status
  echo        v0.0.1           wasmCloud Echo Example                 Compensating
  kvcounter   v0.0.1           wasmCloud Key Value Counter Example      Undeployed
```

emattiza:
## Feature or Problem

updates app listing to be sorted on model name
## Related Issues

closes #156
## Release Information

next
## Consumer Impact

anyone listing models with wadm will now have a default sort applied
## Testing

None added, but happy to implement with the right pointers

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)

None added, but happy to add
### Acceptance or Integration

None added, but happy to add
### Manual Verification

Ran wadm build against sample app and saw a stable sort on repeated invocations

### DIFF
--- a/src/server/storage.rs
+++ b/src/server/storage.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::BTreeSet;
 
 use anyhow::Result;
 use async_nats::jetstream::kv::{Operation, Store};
@@ -130,11 +130,12 @@ impl ModelStorage {
                 }
             });
 
+        // Flatten, collect, and sort on name
         futures::future::join_all(futs)
             .await
             .into_iter()
             .flatten()
-            .collect()
+            .collect::<Result<Vec<ModelSummary>>>()
     }
 
     /// Deletes the given model from storage. This also removes the model from the list of all
@@ -172,7 +173,7 @@ impl ModelStorage {
         &self,
         account_id: Option<&str>,
         lattice_id: &str,
-    ) -> Result<Option<(HashSet<String>, u64)>> {
+    ) -> Result<Option<(BTreeSet<String>, u64)>> {
         match self
             .store
             .entry(model_set_key(account_id, lattice_id))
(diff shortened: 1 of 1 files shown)


src/server/storage.rs
```diff
@@ -130,11 +130,14 @@ impl ModelStorage {
...
 
-        futures::future::join_all(futs)
+        // Flatten, collect, and sort on name
+        let mut models = futures::future::join_all(futs)
             .await
             .into_iter()
             .flatten()
-            .collect()
+            .collect::<Result<Vec<ModelSummary>>>()?;
+        models.sort_by_key(|model| model.name.to_owned());
```
thomastaylor312, line 139: Nit: Can you make this return a borrowed string rather than cloning? With large lists of models this could become rather inefficient. If we have to return an owned key, it might be better to use sort_by_cached_key

thomastaylor312 (dismissed):
Thanks for doing this! I saw this again the other day and it was bugging me. I’ll leave this open for a bit in case you’d like to address the nit, otherwise I can do that in follow up!

src/server/storage.rs
emattiza: i may have to go a bit higher up and tweak some core data choices and implement a btreeset ipo hash set from get_model_set. which will do an initial ordering on String instead of post-hoc ordering in a vec and cloning.

thomastaylor312: Yeah that is actually a really nice solution!

thomastaylor312 (approved):

Issue #154, opened 2023-08-14
[FEAT] Assign leftover instances by weight, instead of spread order


brooksmtownsend:
### Summary
In the spreadscaler, the [compute_spread](https://github.com/wasmCloud/wadm/blob/main/src/scaler/spreadscaler/mod.rs#L345) function takes a spreadscaler configuration and returns a list of tuples of the spread and the total number of replicas to assign to that spread. This is generally pretty fair, taking into account weight and the requested number of replicas in order to fully spread that resource (actor or provider) across the different hosts.

### Current behavior
However, because of part of the algorithm, we end up slightly favoring spread requirements defined earlier in the YAML, rather than the ones with the highest weight. Consider this code, the first line calculates the number of replicas to assign to a spread and the second iteration will take the remaining replicas after that calculation and assign them in the order that the iter returns:
```rust
                // Order is important here since usizes chop off remaining decimals
                (replicas * s.weight.unwrap_or(DEFAULT_SPREAD_WEIGHT)) / total_weight,
/// elided
            // Take the remainder of replicas and evenly distribute among remaining spreads
            let mut diff = replicas - total_replicas;
            spreads
                .into_iter()
                .map(|(spread, count)| {
                    let additional = if diff > 0 {
                        diff -= 1;
                        1
                    } else {
                        0
                    };
                    (spread, count + additional)
                })
                .collect()
```

There's nothing wrong with this and, with the snippet after to assign leftover instances, this will always spread the replicas around properly. However, consider the following spread:
```yaml
        - type: spreadscaler
          properties:
            replicas: 1
            spread:
              - name: eastcoast
                requirements:
                  zone: us-east-1
                weight: 20
              - name: westcoast
                requirements:
                  zone: us-west-1
                weight: 80
```

In this scenario, we want to run one instance of the resource with a 20% weight in east, and a 80% weight in west. Based on the current algorithm, we would assign `(1 * .20) / 1 = 0` replicas to the first, and `(1 * .8) / 1 = 0` replicas to the second. So, we'll have one leftover replica, which we will just assign to **eastcoast** because it comes first in the vector of spreads. It's also possible that the list isn't guaranteed to be sorted in that way, which would be worse as it leads to undetermined behavior.

### Expected behavior
The spread above should favor the **westcoast** spread requirement as it has a higher weight, and ignore the order of the YAML list. This is likely a simple change, simply sorting the spread requirements by weight before randomly assigning the remaining replicas.

PR #173, opened 2023-09-12
feat(spreadscaler): Assign leftover instances by weight, instead of spread order

ahmedtadde:
## Feature or Problem
The existing implementation of [compute_spread](https://github.com/wasmCloud/wadm/blob/1bc07979d0e09ff05e4a5a093d785b6f430fb67e/src/scaler/spreadscaler/mod.rs#L349) favors spread requirements defined earlier in the YAML, rather than the ones with the highest weight. With the updated implementation, leftover replicas are distributed based on weight.

## Related Issues
Addresses [#154](https://github.com/wasmCloud/wadm/issues/154)

## Release Information
v0.6.0

## Consumer Impact
N/A

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
The following tests are modified to align with the new implementation where order by weight does matter.
- [can_spread_properly](https://github.com/wasmCloud/wadm/blob/1bc07979d0e09ff05e4a5a093d785b6f430fb67e/src/scaler/spreadscaler/mod.rs#L438)
- [can_compute_spread_commands](https://github.com/wasmCloud/wadm/blob/1bc07979d0e09ff05e4a5a093d785b6f430fb67e/src/scaler/spreadscaler/mod.rs#L587)

### DIFF
--- a/src/scaler/spreadscaler/mod.rs
+++ b/src/scaler/spreadscaler/mod.rs
@@ -1,4 +1,4 @@
-use std::{cmp::Ordering, collections::HashMap};
+use std::{cmp::Ordering, cmp::Reverse, collections::HashMap};
 
 use anyhow::Result;
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


src/scaler/spreadscaler/mod.rs
```diff
@@ -394,10 +399,33 @@ fn compute_spread(spread_config: &SpreadScalerProperty) -> Vec<(Spread, usize)>
...
+        .collect();
+
+    computed_spreads.iter().fold(
+        vec![(Spread::default(), 0); computed_spreads.len()],
+        |mut output, spread| {
+            let idx = requested_spreads_indices.get(&spread.0.name).unwrap();
+            output[*idx] = spread.to_owned();
+            output
+        },
+    )
```
brooksmtownsend, line 428: I think I agree with your comment here that the spread shouldn't matter, and we probably don't need this extra processing step. This is only an internal method to the scaler and won't modify the order that we store the manifest in the kvstore, which is to say the user won't know we're doing this computation.

If you'd like to keep it, the only thing I would request is that the last `fold` is made safer without the use of an `unwrap` and direct Vec indexing 🙂

brooksmtownsend (changes requested):
This is an elegant fix! I have one requested change around re-sorting the computed spreads, leaning towards not needing to do it at all.

src/scaler/spreadscaler/mod.rs
ahmedtadde: fair point. my intuition is that the (re)ordering should be kept at least for the time being. Ergo, i will address the potentially unsafe stuff you mentioned. good call

src/scaler/spreadscaler/mod.rs
```diff
@@ -394,10 +400,46 @@ fn compute_spread(spread_config: &SpreadScalerProperty) -> Vec<(Spread, usize)>
...
+
+    // case when spread_config.spread.is_empty() <=> requested_spreads.is_empty()
+    if computed_spreads.len() == 1 {
+        return computed_spreads;
+    }
+
+    // this is an invariant check; the predicate should never be true.
+    // this check implicitly ensures that the reordering logic below is safe with regards to indexing.
+    if computed_spreads.len() != requested_spreads.len() {
+        panic!("Computed spreads and requested spreads are not the same length. This should not happen");
```
brooksmtownsend, line 416: I would err on removing this check or setting this as a warning log, see below comment re: panic philosophy

src/scaler/spreadscaler/mod.rs
```diff
@@ -394,10 +400,46 @@ fn compute_spread(spread_config: &SpreadScalerProperty) -> Vec<(Spread, usize)>
...
+    computed_spreads.iter().fold(
+        vec![(Spread::default(), 0); computed_spreads.len()],
+        |mut output, spread| {
+            if let Some(idx) = requested_spreads_indices.get(&spread.0.name) {
+                *output.index_mut(*idx) = spread.to_owned()
+            } else {
+                panic!(
+                    "Spread(name={}) was not found in spread_config (by way of requested_spreads_indices). This should not happen",
+                    spread.0.name
+                );
```
brooksmtownsend, line 437: This seems impossible to reach, but I would still rather this be an error log than an explicit `panic` to avoid crashing our process/task in an extreme edge case. FWIW, I avoid any panics/unwraps outside of tests because I never want to see a process crash.

I still feel this bit of logic isn't necessary since this is an internal function to the spreadscaler itself, is your concern that we might want to read or examine the computed spreads to match it up to the YAML here and, when we sort by weight, it might not be the same order as it's specified in the manifest?

brooksmtownsend (changes requested):

src/scaler/spreadscaler/mod.rs
ahmedtadde: yes, in a nutshell, that is my concern. basically, not knowing enough about the entire codebase, i guess i am somewhat wary of subtle bugs that can emerge from other downstream operations that implicitly/unintentionally rely on the order in the YAML.

Now, taking a step back, I am also realizing that your understanding/experience with the codebase should be enough of a reason for me downgrade my concern if you believe returning the list sorted by weight should be fine. With that said, I will go ahead and remove the reordering logic. Thanks for feedback on this by the way.

brooksmtownsend: @ahmedtadde This part of your logic showed some real thought behind it and I really appreciate this part of the contribution. Though it's not present today, feel free to leave a comment noting the change in behavior here and that it would need to be revisited if we depend on a specific order in the future. 

Once your next commit comes in removing this and the previous possible (I mean, probably impossible but just presence of) `panic`, this should be good to go!

brooksmtownsend (approved):

brooksmtownsend:
Thank you @ahmedtadde 🎉

PR #181, opened 2023-09-20
feat!(state): Updates event loop to store state from ActorsStarted/Stopped

thomastaylor312:
This should significantly decrease the amount of writes and reads from the store. Note that I also changed annotations to be a BTreeMap so they could be hashable


connorsmith256 (approved):
LGTM 👍

brooksmtownsend (approved):
🎉

Issue #179, opened 2023-09-19
[FEAT] Validate that linkdef targets actually exist in the manifest


brooksmtownsend:
Today it's totally valid to define a linkdef with a target that's not specified. Because this is a no-op we should reject this at put time.

For example, specifying a trait
```yaml
        - type: linkdef
          properties:
            target: httpserver-notreal
```

When there is no component `httpserver-not-real` should be rejected when the application is put, giving some error message that a linkdef target cannot be found.

brooksmtownsend:
Closed by #182

PR #182, opened 2023-09-21
Validate that linkdef targets actually exist in the manifest

ahmedtadde:
## Feature or Problem
Today it's totally valid to define a `linkdef` with a target that's not specified.
```yaml
        - type: linkdef
          properties:
            target: httpserver-notreal
```
When there is no capability component with name `httpserver-not-real`, manifest validation should return an appropriate error.

## Related Issues
#179

## Release Information
v0.6.0

## Consumer Impact
better DX

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
[test_manifest_validation](https://github.com/ahmedtadde/wadm/blob/786ca22424b1b4cc8b1ec192c88e85baa873edd7/src/server/handlers.rs#L936) was updated to include ensure all declared linkdef targets can be mapped to some capability component.

### DIFF
--- a/src/server/handlers.rs
+++ b/src/server/handlers.rs
@@ -809,14 +809,15 @@ impl<P: Publisher> Handler<P> {
 
 // Manifest validation
 pub(crate) fn validate_manifest(manifest: Manifest) -> anyhow::Result<()> {
-    let mut component_details: HashSet<String> = HashSet::new();
+    let mut name_registry: HashSet<String> = HashSet::new();
+    let mut required_capability_components: HashSet<String> = HashSet::new();
 
     // Map of link names to a vector of provider references with that link name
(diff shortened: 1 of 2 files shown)


thomastaylor312 (approved):
This is excellent @ahmedtadde!

thomastaylor312:
@ahmedtadde So the e2e tests can be a bit flakey, trying to see if there is an actual failure or if it is flaking. Some of the results were definitely flakes, other ones maybe not. But you didn't touch much that should have affected them

Issue #159, opened 2023-08-16
[FEAT] Support upgrades of just link config or provider config


brooksmtownsend:
Currently, the check to ensure that a scaler is meaningfully different from a previous version is done by comparing the IDs to each other. However, the ID for a link scaler does not include its values, and the ID for a provider scaler does not include its provider config. We should generate some form of short hash of linkdef values and provider and append that to the ID so we can delete + put a link, or stop + start a provider when the configuration changes.

ahmedtadde:
hey @brooksmtownsend , i created a draft PR that has changes to address this issue for the `SpreadScaler > LinkScaler`. It is a p.o.c and i want to get some feedback to know if this is in alignment with what you have in mind. If yes, I can go ahead and do the same for the `SpreadScaler > ProviderScaler` (and which other ones?). If not, a pointer or two to set me on the right path would be great.

brooksmtownsend:
@ahmedtadde left some feedback, I think it's exactly what I had in mind! I think the same should be done for the provider SpreadScaler and the provider DaemonScaler since they are both in charge of starting providers. Your approach with `values` should effectively be the same as `config` there to get the hash.

Once you get that updated I think we should look into adding some extra cases to the e2e upgrade test where we test _only_ modifying a link value and a provider config and assert that worked. If you run into any test flakes let me know, we've been trying to improve them lately. As far as I could see with my testing, e2e_multitenant sometimes fails on a status assertion but the core logic works.

ahmedtadde:
hey @brooksmtownsend i pushed several commits for the provider config hash (both, `spreadscaler & daemonscaler`). I also updated `e2e_upgrade` ( let me know if the added test code actually does enough to be useful with regards to the code changes).

connorsmith256:
@brooksmtownsend should this be closed?

PR #185, opened 2023-09-23
[FEAT] Support upgrades of just link config or provider config

ahmedtadde:
## Feature or Problem
Currently, the check to ensure that a scaler is meaningfully different from a previous version is done by comparing the IDs to each other. However, the ID for a link scaler does not include its values, and the ID for a provider scaler does not include its provider config. We should generate some form of short hash of linkdef values and provider and append that to the ID so we can delete + put a link, or stop + start a provider when the configuration changes.

## Related Issues
#159 

## Release Information
v0.7.0

## Consumer Impact
improved correctness for `wadm` manifest change detection

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -1,4 +1,7 @@
-use std::collections::{BTreeMap, HashMap};
+use std::{
+    collections::{BTreeMap, HashMap},
+    hash::{Hash, Hasher},
+};
 
 use serde::{
     de::{self, Deserializer, MapAccess, Visitor},
@@ -179,6 +182,21 @@ impl Serialize for CapabilityConfig {
         match self {
             CapabilityConfig::Json(v) => v.serialize(serializer),
             CapabilityConfig::Opaque(v) => serializer.serialize_str(v),
+        }
+    }
+}
+
+impl Hash for CapabilityConfig {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        match self {
+            CapabilityConfig::Json(v) => {
+                //NOTE(@ahmedtadde): By default, Serde uses BtreeMap for JSON map objects which is good for determinism.
+                //However, for arrays, ordering is not guaranteed.
+                //In general, unless the json values are strictly equal (ordering, casing, white spaces, etc...), the hash will be different. This is something to be mindful of when using this hash function.
+                let json = serde_json::to_string(v).expect("Should be able to serialize");
+                json.hash(state);
+            }
+            CapabilityConfig::Opaque(v) => v.hash(state),
         }
     }
 }
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -193,7 +193,15 @@ impl Hash for CapabilityConfig {
                 //NOTE(@ahmedtadde): By default, Serde uses BtreeMap for JSON map objects which is good for determinism.
                 //However, for arrays, ordering is not guaranteed.
                 //In general, unless the json values are strictly equal (ordering, casing, white spaces, etc...), the hash will be different. This is something to be mindful of when using this hash function.
-                let json = serde_json::to_string(v).expect("Should be able to serialize");
+                let json = match serde_json::to_string(v) {
+                    Ok(json) => json,
+                    Err(e) => {
+                        // This really shouldn't happen, but if it does, we should log it and fallback to a default value '{}' to avoid panicking
+                        eprintln!("Failed to serialize json value: {}", e);
+                        "{}".to_string()
+                    }
+                };
+
                 json.hash(state);
             }
             CapabilityConfig::Opaque(v) => v.hash(state),
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -1,13 +1,13 @@
-use std::{
-    collections::{BTreeMap, HashMap},
-    hash::{Hash, Hasher},
-};
+use std::collections::{BTreeMap, HashMap};
 
+use anyhow::Context;
 use serde::{
     de::{self, Deserializer, MapAccess, Visitor},
     ser::Serializer,
     Deserialize, Serialize,
 };
+
+use base64::{engine::general_purpose, Engine as _};
 
 pub(crate) mod internal;
 
@@ -140,6 +140,25 @@ pub enum CapabilityConfig {
     Opaque(String),
 }
 
+impl CapabilityConfig {
+    pub fn try_base64_encoding(&self) -> anyhow::Result<String> {
+        let mut bytes: Vec<u8> = Vec::new();
+        serde_json::to_writer(&mut bytes, &self)
+            .context("failed to serialize capability config into bytes for base64 encoding")?;
+        Ok(general_purpose::STANDARD.encode(&bytes))
+    }
+
+    pub fn try_base64_encoding_with_engine<T: base64::Engine>(
+        &self,
+        engine: &T,
+    ) -> anyhow::Result<String> {
+        let mut bytes: Vec<u8> = Vec::new();
+        serde_json::to_writer(&mut bytes, &self)
+            .context("failed to serialize capability config into bytes for base64 encoding")?;
+        Ok(engine.encode(&bytes))
+    }
+}
+
 impl<'de> Deserialize<'de> for CapabilityConfig {
     fn deserialize<D>(deserializer: D) -> Result<CapabilityConfig, D::Error>
     where
@@ -185,30 +204,6 @@ impl Serialize for CapabilityConfig {
         }
     }
 }
-
-impl Hash for CapabilityConfig {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        match self {
-            CapabilityConfig::Json(v) => {
-                //NOTE(@ahmedtadde): By default, Serde uses BtreeMap for JSON map objects which is good for determinism.
-                //However, for arrays, ordering is not guaranteed.
-                //In general, unless the json values are strictly equal (ordering, casing, white spaces, etc...), the hash will be different. This is something to be mindful of when using this hash function.
-                let json = match serde_json::to_string(v) {
-                    Ok(json) => json,
-                    Err(e) => {
-                        // This really shouldn't happen, but if it does, we should log it and fallback to a default value '{}' to avoid panicking
-                        eprintln!("Failed to serialize json value: {}", e);
-                        "{}".to_string()
-                    }
-                };
-
-                json.hash(state);
-            }
-            CapabilityConfig::Opaque(v) => v.hash(state),
-        }
-    }
-}
-
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 pub struct Trait {
     /// The type of trait specified. This should be a unique string for the type of scaler. As we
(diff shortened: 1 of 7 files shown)


src/scaler/spreadscaler/link.rs
```diff
@@ -285,6 +295,12 @@ impl<S: ReadStore + Send + Sync, L: LinkSource> LinkScaler<S, L> {
     }
 }
 
+fn compute_linkscaler_values_hash(values: &HashMap<String, String>) -> u64 {
+    let mut linkscaler_values_hasher = std::collections::hash_map::DefaultHasher::new();
+    BTreeMap::from_iter(values.iter()).hash(&mut linkscaler_values_hasher);
+    linkscaler_values_hasher.finish()
+}
```
brooksmtownsend, line 302: @thomastaylor312 you've done more in-depth hashing things in Rust than I, does this look like a good approach to getting a hash of the map?

brooksmtownsend (commented):
So far I like this approach. Setting scalers to have unique identifiers that can determine if a "new" scaler is functionally different lets what we have in place for upgrades work. @thomastaylor312 do you have any concerns with this, e.g. the ID getting too large? Or displaying the hash in a place that a bad actor could decode it and get the values inside?

This supports upgrades well IMO and doesn't fall into the same edge case of competing link scaler config 👍🏻

src/scaler/spreadscaler/link.rs
```diff
@@ -219,7 +222,14 @@ impl<S: ReadStore + Send + Sync, L: LinkSource> LinkScaler<S, L> {
...
         // NOTE(thomastaylor312): Yep, this is gnarly, but it was all the information that would be
         // useful to have if uniquely identifying a link scaler
-        let id = format!("{LINK_SCALER_TYPE}-{model_name}-{provider_link_name}-{actor_reference}-{provider_reference}");
+        let id = if let Some(linkscaler_values) = &values {
+            // When values are present, we want to include them in the ID as well
+            let linkscaler_values_hash = compute_linkscaler_values_hash(linkscaler_values);
+            format!("{LINK_SCALER_TYPE}-{model_name}-{provider_link_name}-{actor_reference}-{provider_reference}-{linkscaler_values_hash}")
+        } else {
+            format!("{LINK_SCALER_TYPE}-{model_name}-{provider_link_name}-{actor_reference}-{provider_reference}")
+        };
```
brooksmtownsend, line 231: I am a little concerned about this unique ID getting large but I don't know if that would cause issues 🤔

tests/e2e_upgrades.rs
```diff
@@ -298,4 +298,116 @@ async fn test_upgrade(client_info: &ClientInfo) {
         Ok(())
     })
     .await;
+
+    // Deploy another updated manifest -- this time just w/ link values and provider config modifications
+    let resp = client_info
+        .put_manifest_from_file("upgradedapp2.yaml", None, None)
+        .await;
+
```
brooksmtownsend, line 306: ❤️ the additional tests here!

brooksmtownsend (commented):
@thomastaylor312 Requesting you review this one as well since this looks exactly how I imagined implementing it but I want to make sure I'm not missing any assumptions around ID lengths and proper ways to use the hash traits.

src/model/mod.rs
```diff
@@ -183,6 +186,21 @@ impl Serialize for CapabilityConfig {
...
 
+impl Hash for CapabilityConfig {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        match self {
+            CapabilityConfig::Json(v) => {
+                //NOTE(@ahmedtadde): By default, Serde uses BtreeMap for JSON map objects which is good for determinism.
+                //However, for arrays, ordering is not guaranteed.
+                //In general, unless the json values are strictly equal (ordering, casing, white spaces, etc...), the hash will be different. This is something to be mindful of when using this hash function.
+                let json = serde_json::to_string(v).expect("Should be able to serialize");
+                json.hash(state);
```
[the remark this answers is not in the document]
ahmedtadde: in terms of fallback behavior, if the serialization fails, should the hash prefixing just be skipped OR should an error be returned as opposed to a full blown `panic`?

slightly related...is there an enforced size limit on the `provider_config` object? If not, should there be? i guess the same questions applies to the linkdef `values`. `"we'll deal with it if it becomes a problem"` is also a reasonable approach for now, I suppose.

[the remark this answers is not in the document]
brooksmtownsend: I don't know if you can return an error to hash, so it may just be good fallback behavior to default to some other value. 

For your related question, there isn't an enforced size limit so that may be an issue. We may want to use a hash method that uses a short hash value since we don't need to go from hash to values, we just want a uniqueness summary 🤔

src/scaler/spreadscaler/link.rs
thomastaylor312: I think for the purposes we're using it for here, this works great

src/scaler/spreadscaler/link.rs
ahmedtadde: may be a separate ticket for this (if/when it becomes a problem)?

thomastaylor312:
@ahmedtadde I will be getting to this I promise! Just dealing with some other bugs in wadm right now and trying to get those handled

src/scaler/spreadscaler/link.rs
thomastaylor312: We could always start hashing this text if we need to. But I'm ok to yolo for now

ahmedtadde:
> @ahmedtadde I will be getting to this I promise! Just dealing with some other bugs in wadm right now and trying to get those handled

No problem!

src/scaler/daemonscaler/provider.rs
```diff
@@ -209,10 +210,20 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ProviderDaemonScaler<S> {
 impl<S: ReadStore + Send + Sync> ProviderDaemonScaler<S> {
     /// Construct a new ProviderDaemonScaler with specified configuration values
     pub fn new(store: S, config: ProviderSpreadConfig, component_name: &str) -> Self {
-        let id = format!(
-            "{PROVIDER_DAEMON_SCALER_TYPE}-{}-{component_name}-{}-{}",
-            config.model_name, config.provider_reference, config.provider_link_name,
-        );
+        let id = if let Some(provider_config) = &config.provider_config {
```
thomastaylor312, line 213: So thinking back on this, can you walk me though why we want to hash the provider config here? Within wasmcloud, the link name uniquely identifies the provider. In other words, there isn't a world where a provider with link name `foo` and config `{"blah": 2}` can exist at the same time as the same provider with link name `foo` and config `{"bar": "baz"}`.

src/model/mod.rs
```diff
@@ -183,6 +186,29 @@ impl Serialize for CapabilityConfig {
...
 }
 
+impl Hash for CapabilityConfig {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        match self {
+            CapabilityConfig::Json(v) => {
+                //NOTE(@ahmedtadde): By default, Serde uses BtreeMap for JSON map objects which is good for determinism.
+                //However, for arrays, ordering is not guaranteed.
+                //In general, unless the json values are strictly equal (ordering, casing, white spaces, etc...), the hash will be different. This is something to be mindful of when using this hash function.
+                let json = match serde_json::to_string(v) {
```
thomastaylor312, line 196: If we do need to keep this (see other comment), I worry about having to serialize every time we want to hash. I know we only do it once, but it feels like with that overhead it doesn't belong in a `Hash` implementation so much as written into something the a base64 encoder and hashing that way

thomastaylor312 (changes requested):
Really love the link stuff. I do have some comments/questions around the provider config hashing

src/scaler/daemonscaler/provider.rs
ahmedtadde: good question. 

so, assuming that `provider_config` does impact how the scaler/wadm handles a provider on the lattice, [my understanding](https://github.com/wasmCloud/wadm/issues/159) is that we want  a manifest change  `foo w/ {"blah": 2} => foo w/ {"blah": "baz" }` to trigger some steps that the scaler/wadm needs to take for a desired state. Basically, i was under the impression that, as things currently stand, `foo w/ {"blah": 2} => foo w/ {"blah": "baz" }`  doesn't instigate any state transition despite because a change of `provider_config: {"blah": 2}  =>  {"blah": "baz" }` is not reflected in the current  change detection implementation.

@brooksmtownsend  let me know if I misunderstood the intent of #159 with regards to `provider_config`

src/model/mod.rs
ahmedtadde: [is this change inline with what you have in mind](https://github.com/wasmCloud/wadm/pull/185/commits/cab231f5402b1b9c95e16e8925b7f3ac0127ea64)

src/scaler/daemonscaler/provider.rs
```diff
@@ -258,6 +281,17 @@ impl<S: ReadStore + Send + Sync> ProviderDaemonScaler<S> {
     }
 }
 
+fn compute_provider_config_hash(provider_config: &CapabilityConfig) -> Option<u64> {
+    match provider_config.try_base64_encoding() {
+        Ok(base64) => {
+            let mut hasher = std::collections::hash_map::DefaultHasher::new();
```
thomastaylor312, line 287: I think we just have the hash be a string rather than hashing again into a number. So instead of this function, you can just do `provider_config.try_base64_encoding().ok()` instead

thomastaylor312 (changes requested):
Yep, I think I like the base64 option better here. Just one more minor change and then this should be good to go

src/scaler/daemonscaler/provider.rs
ahmedtadde: cool cool

thomastaylor312 (dismissed):

thomastaylor312:
Hey @ahmedtadde sorry to come back again, but I just noticed you have some left over imports that would be good to remove:

```
warning: unused import: `Hash`
 --> src/scaler/daemonscaler/provider.rs:2:17
  |
2 | use std::hash::{Hash, Hasher};
  |                 ^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `Hasher`
 --> src/scaler/daemonscaler/provider.rs:2:23
  |
2 | use std::hash::{Hash, Hasher};
  |                       ^^^^^^

warning: unused import: `Hash`
 --> src/scaler/spreadscaler/provider.rs:4:12
  |
4 |     hash::{Hash, Hasher},
  |            ^^^^

warning: unused import: `Hasher`
 --> src/scaler/spreadscaler/provider.rs:4:18
  |
4 |     hash::{Hash, Hasher},
  |                  ^^^^^^
```

Also, I just merged my PR that should help fix some of the flaky tests, so if you want to rebase while you're at it, it might make things go smoother. Otherwise, I'm about to cut the 0.8.0-rc1, so I can also fix that in my follow up if you don't have time

ahmedtadde:
@thomastaylor312 fixed the imports and rebased

thomastaylor312:
@ahmedtadde It looks like you ended up merging rather than rebasing? Something looks funky because now the PR has all my changes showing up as diffs too 😅

thomastaylor312:
If you don't want to mess with it, I can rebase and force push to your branch, but only if you want me to. I feel bad this went sideways on you

ahmedtadde:
> @ahmedtadde It looks like you ended up merging rather than rebasing? Something looks funky because now the PR has all my changes showing up as diffs too 😅

ah sh***t, my bad! I done did it! yeah, you can force push

thomastaylor312 (approved):

Issue #180, opened 2023-09-20
[FEAT] Add the ability to set a stream prefix for all wadm internal streams


thomastaylor312:
When running in a multitenant environment, clients will need to share the same JS domain (since messages need to come from lattices). However, this means you can't have a separate stream for different wadm's running in a multitenant environment (e.g. `wadm_status`).

I think it would be good to allow a user to configure a stream prefix that would be added to each stream: `<myprefix>_wadm_status`

ahmedtadde:
Hello @thomastaylor312 , Assuming this still available to be worked on (and still relevant), i'd like to take a stab at it. so far, i have not been able to pin point the place(s) in the code to make this change. I looked at `nats_utils`, `lib` (ie, the constant prefixes in there), `workers/*`, `consumers/*` and `events/*` and I could readily determine where the change will need to be made. I'd appreciate a push in the right direction; maybe there is some context that I am missing.

thomastaylor312:
Sorry for the delay @ahmedtadde. So this is actually in the `bin/` directory. We have a bunch of `const`s we have there right now: https://github.com/wasmCloud/wadm/blob/867793f871f80d038ba7070abc1281ae4748b034/bin/main.rs#L3. So you'd add a new flag that would allow you to set a prefix and use it in conjunction with those constants

brooksmtownsend:
closed by #188

PR #188, opened 2023-10-06
[FEAT] Add the ability to set a stream prefix for all wadm internal streams 

ahmedtadde:
## Feature or Problem
 currently, one can't have a separate stream for different wadms' running in a multitenant environment (e.g. wadm_status).

## Related Issues
#180

## Release Information
v0.7.x

## Consumer Impact
adds some flexibility to `wadm`

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/bin/main.rs
+++ b/bin/main.rs
@@ -137,8 +137,10 @@ struct Args {
     )]
     api_prefix: String,
 
-    /// The Stream topic prefix to use. This is an advanced setting that should only be used if you
-    /// know what you are doing
+    /// This prefix to used for the internal streams. When running in a multitenant environment,
+    /// clients share the same JS domain (since messages need to come from lattices).
+    /// Setting a stream prefix makes it possible to have a separate stream for different wadms running in a multitenant environment.
+    /// This is an advanced setting that should only be used if you know what you are doing.
     #[arg(long = "stream-prefix", env = "WADM_STREAM_PREFIX")]
     stream_prefix: Option<String>,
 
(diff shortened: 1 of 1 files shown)


bin/main.rs
```diff
@@ -137,6 +137,11 @@ struct Args {
     )]
     api_prefix: String,
 
+    /// The Stream topic prefix to use. This is an advanced setting that should only be used if you
```
thomastaylor312, line 140: Can you add just a bit more documentation here? Something that says this sets the prefix for all the internal streams so you can run different wadms within the same environment?

thomastaylor312 (changes requested):
One super minor change and this should be ready to go. Thanks again here @ahmedtadde!

bin/main.rs
ahmedtadde: yep. no problem. I updated the comment to reflect the intended use based on the [#180](https://github.com/wasmCloud/wadm/issues/180) description. let me know if it needs some improvement.

thomastaylor312 (approved):

bin/main.rs
thomastaylor312: Nope, looks good. Thank you!

PR #190, opened 2023-10-12
fix(*): Link caching and other goodies

thomastaylor312:
This PR got a little bit out of hand. We had a bunch of bugs that may or may not have been interrelated so we fixed most of them. The biggest thing is that we switching to caching link definitions for scalers so that the number of consumers created was much lower. Additionally, we fixed when status notifications are published so that we don't get reconciling status when other scalers are being removed


bin/main.rs
```diff
@@ -173,14 +174,23 @@ async fn main() -> anyhow::Result<()> {
...
         client.clone(),
         ControlClientConfig {
-            js_domain: args.domain,
+            js_domain: args.domain.clone(),
             topic_prefix: None,
         },
     );
 
+    let hack_connection_pool: ControlClientConstructor<DirectKvStore> =
+        ControlClientConstructor::new(
```
brooksmtownsend, line 186: I think this is leftover from testing

bin/main.rs
```diff
@@ -272,7 +282,7 @@ async fn main() -> anyhow::Result<()> {
     debug!("Creating command consumer manager");
 
     let command_worker_creator = CommandWorkerCreator {
-        pool: connection_pool,
+        pool: hack_connection_pool,
```
brooksmtownsend, line 285:
```suggestion
        pool: connection_pool,
```

bin/main.rs
```diff
@@ -333,12 +343,12 @@ async fn main() -> anyhow::Result<()> {
...
 #[derive(Clone)]
 struct CommandWorkerCreator {
-    pool: ControlClientConstructor,
+    pool: ControlClientConstructor<DirectKvStore>,
 }
 
 #[async_trait::async_trait]
 impl WorkerCreator for CommandWorkerCreator {
-    type Output = CommandWorker;
+    type Output = CommandWorker<DirectKvStore>;
```
brooksmtownsend, line 351: Is this intended to be directly querying or does it just not matter since for commands we're just publishing to the KV bucket?

src/scaler/manager.rs
```diff
@@ -416,8 +431,16 @@ where
                                 Notifications::DeleteScalers(name) => {
                                     trace!(%name, "Removing scalers for manifest");
                                     match self.remove_scalers_internal(&name).await {
-                                        Some(Ok(_)) => {
-                                            trace!(%name, "Successfully removed scalers for manifest")
+                                        Some(Ok(_)) | None => {
+                                            trace!(%name, "Removed manifests or manifests were already removed");
+                                            // NOTE(thomastaylor312): We
```
brooksmtownsend, line 436: We what 👀 😮

src/workers/event.rs
```diff
@@ -1427,11 +1430,17 @@ mod test {
                                     annotations: None,
                                     instance_id: "1".to_string(),
                                     revision: 0,
+                                    // TODO: FIXME
+                                    image_ref: None,
+                                    max_concurrent: 1,
```
brooksmtownsend, line 1435: For all of these, leaving the image_ref as `None` and max_concurrent as 0 simulates backwards-compatible behavior. I think we should do that until we officially update to v0.79.0 and use these fields for realsies

brooksmtownsend (changes requested):
By and large looks great just a couple nit requests, the only other thing I'd like to include here is updating the control interface to handle host inventories with the "new" instance representation and the old. I'll follow up with you on this today 👍🏻

bin/main.rs
thomastaylor312: Actually it is meant to be there, but I didn't change the name from testing. Will change now

brooksmtownsend (approved):
🙌🏻

PR #193, opened 2023-10-12
Modified provider checks to be version independant

aish-where-ya:
## Feature or Problem
Bug: Providers deployed in a manifest on a link name needed to have a unique image reference associated with it. This allowed for providers with the same reference but different versions to be created on the same link name. 

Fix: This PR improves upon this issue by associating an image ref, regardless of the version, with a link name. By ignoring the version, we permit only unique image references into the manifest/lattice.

## Release Information
next

## Consumer Impact
Capabilities with duplicate image refs will be removed.

## Testing
manually tested

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Modified `test_manifest_validation`

### Acceptance or Integration
Updated e2e test named `e2e_upgrades` to check for provider versioning

### Manual Verification
manually verified.

### DIFF
--- a/src/server/handlers.rs
+++ b/src/server/handlers.rs
@@ -521,6 +521,91 @@ impl<P: Publisher> Handler<P> {
                 }
             };
 
+        // Retrieve all stored models in the lattice
+        let stored_models = match self.store.list(account_id, lattice_id).await {
+            Ok(d) => d,
+            Err(e) => {
+                error!(error = %e, "Unable to fetch data");
+                self.send_error(msg.reply, "Internal storage error".to_string())
+                    .await;
+                return;
+            }
+        };
+
+        // Retrieve all the existing provider refs in store that are currently deployed
+        let mut existing_provider_refs: HashSet<String> = HashSet::new();
+        for model_summary in stored_models.iter() {
+            if let Some(deployed_version) = &model_summary.deployed_version {
+                let (stored_manifest, _) = match self
+                    .store
+                    .get(account_id, lattice_id, &model_summary.name)
+                    .await
+                {
+                    Ok(Some(m)) => m,
+                    Ok(None) => (StoredManifest::default(), 0),
+                    Err(e) => {
+                        error!(error = %e, "Unable to fetch data");
+                        self.send_error(msg.reply, "Internal storage error".to_string())
+                            .await;
+                        return;
+                    }
+                };
+                if let Some(deployed_manifest) = stored_manifest.get_version(deployed_version) {
+                    for component in deployed_manifest.spec.components.iter() {
+                        if let Properties::Capability {
+                            properties:
+                                CapabilityProperties {
+                                    image: image_name, ..
+                                },
+                        } = &component.properties
+                        {
+                            let image_ref_split: Vec<&str> = image_name.split(':').collect();
+                            if let Some(&ref_link) = image_ref_split.first() {
+                                existing_provider_refs.insert(ref_link.to_string());
+                            }
+                        }
+                    }
+                };
+            }
+        }
+
+        let staged_model = match req.version.clone() {
+            Some(v) => {
+                if let Some(model) = manifests.get_version(&v) {
+                    model
+                } else {
+                    manifests.get_current()
+                }
+            }
+            None => manifests.get_current(),
+        };
+
+        // Compare if any of the provider refs in the staged model are duplicates
+        for component in staged_model.spec.components.iter() {
+            if let Properties::Capability {
+                properties:
+                    CapabilityProperties {
+                        image: image_name, ..
+                    },
+            } = &component.properties
+            {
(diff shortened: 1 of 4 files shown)


brooksmtownsend (commented):
This does explicitly block people from pushing different providers to the same image repository, but I think that's an okay assertion 🤔 

Thoughts @thomastaylor312?

aish-where-ya:
We can always pick the other route and explicitly perform checks against the version. It would involve retrieving items from the linkedef hash map, splitting it into reference and version and performing checks for this specific case where the user is trying to put in an existing provider with a different version. 

I'm thinking about cases where the approach in the PR will fail/succeed to catch what the other approach can/cannot. This one is a lesser number of steps that's all.

connorsmith256:
Does wadm have access to the lattice's metadata KV bucket? If so, wadm should be able to translate from the providers' image ref to its public key, which might be a better test for uniqueness

autodidaddict:
> Does wadm have access to the lattice's metadata KV bucket? If so, wadm should be able to translate from the providers' image ref to its public key, which might be a better test for uniqueness

The image ref to public key lookup can be gleaned from making a control interface API request for the claims cache. Giving wadm direct access to buckets like those of  the lattice cache can be potentially problematic in some production scenarios, especially multi-tenant and high security.

src/server/handlers.rs
```diff
@@ -521,6 +521,102 @@ impl<P: Publisher> Handler<P> {
...
+                    Ok(None) => (StoredManifest::default(), 0),
+                    Err(e) => {
+                        error!(error = %e, "Unable to fetch data");
+                        self.send_error(msg.reply, "Internal storage error".to_string())
+                            .await;
+                        return;
+                    }
+                };
+                if stored_manifest.name() != name {
+                    if let Some(deployed_manifest) = stored_manifest.get_version(deployed_version) {
```
brooksmtownsend, line 554: Consider using `stored_manifest.get_deployed` to simplify this loop

src/server/handlers.rs
```diff
@@ -521,6 +521,102 @@ impl<P: Publisher> Handler<P> {
...
+                {
+                    Ok(Some(m)) => m,
+                    Ok(None) => (StoredManifest::default(), 0),
+                    Err(e) => {
+                        error!(error = %e, "Unable to fetch data");
+                        self.send_error(msg.reply, "Internal storage error".to_string())
+                            .await;
+                        return;
+                    }
+                };
```
brooksmtownsend, line 552: I think if you use the suggestion in the above comment with `get_deployed` you can get rid of this check altogether 😄

src/server/handlers.rs
```diff
@@ -521,6 +521,102 @@ impl<P: Publisher> Handler<P> {
...
+            if let Properties::Capability {
+                properties:
+                    CapabilityProperties {
+                        image: image_name, ..
+                    },
+            } = &component.properties
+            {
+                let image_ref_split: Vec<&str> = image_name.split(':').collect();
+                if let (Some(&ref_link), Some(&ref_version)) =
+                    (image_ref_split.first(), image_ref_split.last())
```
brooksmtownsend, line 599: This shows up in a place or two, if it makes sense consider a function like
```
fn parse_image_ref_and_tag(image_name: &str) -> Option<(String, String)>
```
I'm bad at naming but you get the idea 😄

src/server/handlers.rs
```diff
@@ -521,6 +521,102 @@ impl<P: Publisher> Handler<P> {
...
+                {
+                    if let Some(old_version) = existing_provider_refs.get(ref_link) {
+                        if old_version != ref_version {
+                            error!(
+                                "Provider {image_name} is currently deployed. Forbidden operation.",
+                            );
+                            self.send_error(
+                                msg.reply,
+                                format!(
+                                "Provider {image_name} is currently deployed. Forbidden operation."
```
brooksmtownsend, line 609: Is there a way here to indicate which model is deployed with the offending provider?

```suggestion
                                "Provider {image_name} is currently deployed with a different version. Forbidden operation.",
                            );
                            self.send_error(
                                msg.reply,
                                format!(
                                "Provider {image_name} is currently deployed with a different version. Forbidden operation."
```

brooksmtownsend (changes requested):
A couple requests for small efficiency improvements 🙂

src/server/handlers.rs
aish-where-ya: Modifying the check instead of removing since we don't want to perform a fetch operation if there is no currently deployed version.

brooksmtownsend (approved):

Issue #162, opened 2023-08-18
[BUG] Fix README styles with wadm logo


brooksmtownsend:
Pretty funny bug here, but on GitHub the README is wider than crates.io so it looks fine on GitHub. We may need to set a max-width in the README or better scale the logo 
<img width="1058" alt="image" src="https://github.com/wasmCloud/wadm/assets/12040685/dd017bc5-71e9-4d68-907c-9cfc2e6a0f0f">

PR #196, opened 2023-10-18
chore: add smaller logos

connorsmith256:
## Feature or Problem
The logo at https://crates.io/crates/wadm is too large and makes the title difficult to read

This adds two smaller-resolution versions of the wadm logo under `static/images`

Since https://wasmcloud.com/docs/fundamentals/wadm/ references the image at the root of this repo, I left the original there, which we can remove after this is merged and the docs site is updated as well 

## Related Issues
Resolves https://github.com/wasmCloud/wadm/issues/162

## Release Information
N/A

## Consumer Impact
Our users can read our words more easily :)

## Testing
See https://github.com/connorsmith256/wadm/tree/chore/wadm-images#readme for what the result of this change looks like


brooksmtownsend (approved):
Now to find the integration test that depends on a big logo...

Issue #191, opened 2023-10-12
Remove backwards compatible HostInventory logic with wasmCloud 0.80.0


brooksmtownsend:
Between 0.78 and 0.79 the wasmcloud-control-interface updated to change tracking of instances from a list of instances to a single instance that reports its own level of max_concurrency (count). In #190 we added backwards compatible logic to handle both inventory representations but once wasmCloud is two minor versions ahead (0.80.0) we should remove that logic in favor of the simpler representation.

As a part of that PR we tagged this issue in code comments so it's easy to find where these changes have to take place.

PR #199, opened 2023-10-19
Fully reconstitute state from heartbeats

brooksmtownsend:
## Feature or Problem
This PR adds additional logic to the host heartbeat handling to make sure that actors and providers fully reconstitute state from heartbeats, even if the `wadm_state` bucket was entirely deleted. Wadm currently restores _most_ critical information, but notably did not restore the image reference which would require you to delete and restart the actor or provider in order for wadm to find it.

## Related Issues
Mentioned #191 in the code here.

## Release Information
I think we should take out a release branch to publish a patch for 0.7.2 with this commit at least cherry picked in order to not hold this back on 0.8.0.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Updated the heartbeat missing information test to actually test that we restore _all_ data.

### Manual Verification
I tested this both with partial information and an entirely deleted `wadm_state` bucket.

### DIFF
--- a/src/workers/event.rs
+++ b/src/workers/event.rs
@@ -755,6 +755,14 @@ where
                         prov.contract_id = info.contract_id.clone();
                         has_changes = true;
                     }
+                    if prov.name.is_empty() {
+                        prov.name = info.name.clone().unwrap_or_default();
+                        has_changes = true;
+                    }
+                    if prov.reference.is_empty() {
+                        prov.reference = info.image_ref.clone().unwrap_or_default();
+                        has_changes = true;
+                    }
                     if let Entry::Vacant(entry) = prov.hosts.entry(host.id.clone()) {
                         entry.insert(ProviderStatus::default());
                         has_changes = true;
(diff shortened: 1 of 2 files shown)


src/workers/event.rs
```diff
@@ -2367,8 +2422,8 @@ mod test {
                 issuer: "my-issuer-5".to_string(),
                 actors: vec![ActorDescription {
                     id: "jabba".to_string(),
-                    image_ref: None,
-                    name: None,
+                    image_ref: Some("jabba.tattoinecr.io/jabba:latest".to_string()),
```
thomastaylor312, line 2425: I'm sorry sir, it is _ackshually_ tatooine, not tattoine, unless you're looking for a Star Wars themed tattoo shop

thomastaylor312 (approved):

thomastaylor312 (approved):

PR #211, opened 2023-11-02
fix(storage): Updates scalers to do a single data lookup per event

### LINKED ISSUE #203: State lookups use a ton of bandwidth
When running wadm in production at cosmonic, we discovered that wadm uses a ton of bandwidth querying data from the store. Back of napkin math put us around 500GB/month just for querying linkdefs and wadm state data. I will document a more elegant solution in another issue, but we should probably just do a single "snapshot" lookup of state (wadm_state + linkdefs + claims) that is then used by all scalers handling a given event. This means that rather than each scaler querying individually, it will just be queried once.

thomastaylor312:
This works by looking up all the data once per event (whenever scalers are called). This reduces lookups of links by at least 40% and should reduce data usage as well (though not as much as doing some sort of caching would)

Comparing number of `get.links` requests coming over the control topics:

```
❯
rg 'get.links' /tmp/test-e2e-after | wc -l
     272

❯
rg 'get.links' /tmp/test-e2e-before | wc -l
     459
```

Fixes #203

### DIFF
--- a/src/storage/snapshot.rs
+++ b/src/storage/snapshot.rs
@@ -1,0 +1,161 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use tokio::sync::RwLock;
+use wasmcloud_control_interface::LinkDefinition;
+
+use crate::storage::{Actor, Host, Provider, ReadStore, StateKind};
+use crate::workers::LinkSource;
+
+// NOTE(thomastaylor312): This type is real ugly and we should probably find a better way to
+// structure the ReadStore trait so it doesn't have the generic T we have to work around here. This
+// is essentially a map of "state kind" -> map of ID to partially serialized state. I did try to
+// implement some sort of getter trait but it has to be generic across T
+type InMemoryData = HashMap<String, HashMap<String, serde_json::Value>>;
+
+/// A store and claims/links source implementation that contains a static snapshot of the data that
+/// can be refreshed periodically. Please note that this is scoped to a specific lattice ID and
+/// should be constructed separately for each lattice ID.
+///
+/// NOTE: This is a temporary workaround until we get a proper caching store in place
+pub struct SnapshotStore<S, L> {
+    store: S,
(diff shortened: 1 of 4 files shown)


src/storage/snapshot.rs
```diff
@@ -0,0 +1,161 @@
...
+    fn clone(&self) -> Self {
+        Self {
+            store: self.store.clone(),
+            link_source: self.link_source.clone(),
+            lattice_id: self.lattice_id.clone(),
+            stored_state: self.stored_state.clone(),
+            links: self.links.clone(),
+        }
+    }
+}
```
brooksmtownsend, line 43: Do you need this to restrict clone to where S, L are clonable?

src/storage/snapshot.rs
```diff
@@ -0,0 +1,161 @@
...
+            link_source,
+            lattice_id,
+            stored_state: Default::default(),
+            links: Arc::new(RwLock::new(Vec::new())),
+        }
+    }
+
+    /// Refreshes the snapshotted data, returning an error if it couldn't update the data
+    pub async fn refresh(&self) -> anyhow::Result<()> {
+        // SAFETY: All of these unwraps are safe because we _just_ deserialized from JSON
```
brooksmtownsend, line 63: spooky

brooksmtownsend (approved):
Good stuff 👍🏻

src/storage/snapshot.rs
brooksmtownsend: To clarify, I think this code is correct just asking if that's the reason why this isn't `derive(Clone)`

thomastaylor312: This is for cleanliness here mostly. We know that anything we pass in here will be clone, but many times you can’t actually know that. For us here, if we want to derive clone, we’d have to add a clone bound to the top level strict definition. Then that clone bound bleeds to every other impl of the struct

Issue #195, opened 2023-10-17
Validation errors should give useful debugging information


thomastaylor312:
When I was debugging the issue in #194, I was receiving the following validation error:

```
Validation error in object: {"name":"nats","properties":{"contract":"wasmcloud:messaging","image":"wasmcloud.azurecr.io/nats_messaging:0.17.3"},"traits":[{"properties":{"spread":[{"name":"onkube","requirements":{"cosmo_connect":"k8s","kubernetes":"true"}}]},"type":"daemonscaler"}],"type":"capability"}
Object path: /spec/components/1Validation error in object: {"name":"applier","properties":{"contract":"cosmonic:kubernetes_applier","image":"wasmcloud.azurecr.io/applier:0.3.0"},"traits":[{"properties":{"spread":[{"name":"onkube","requirements":{"cosmo_connect":"k8s","kubernetes":"true"}}]},"type":"daemonscaler"}],"type":"capability"}
Object path: /spec/components/2Validation error in object: {"name":"httpserver","properties":{"contract":"wasmcloud:httpserver","image":"wasmcloud.azurecr.io/httpserver:0.19.1"},"traits":[{"properties":{"spread":[{"name":"onkube","requirements":{"cosmo_connect":"k8s","kubernetes":"true"}}]},"type":"daemonscaler"}],"type":"capability"}
Object path: /spec/components/3
```

While this is better than nothing, what it doesn't do is provide any indication as to what is missing or wrong with the data. We should update our errors to be more clear about what is missing. An error above should have said something like

```
Validation error with component "nats": Missing field `replicas`
```

PR #214, opened 2023-11-02
chore: Better JSON validation error messages

aish-where-ya:
## Feature or Problem
Better error messages for JSON validation.

Currently, when manifests are validated, the deserialization and validation is internally  handled by serde and it prints out error messages that specify the exact location of the error in the file. Here are some of the errors thrown by serde:
```
Should be able to parse: spec.components[5]: missing field `contract` at line 97 column 7

Should be able to parse: spec.components[0]: invalid type: unit value, expected struct ActorProperties at line 10 column 7
```

Serde handles a majority of the cases in the `Manifest` schema but in the case of `Traits`,  if there are any missing or incorrect properties that do not fall under type `Linkedef` or `Spreadscaler`, serde deserializes to `Custom` `TraitProperty`.  Upon which the validation falls through to the OAM validation schema. Due to the limited features of the `jsonschema` crate, it pinpoints to the correct area with the problematic code but does not tell the user the exact problem with the schema.

This PR takes the best it can of the `ValidationError` provided by `jsonschema` and gives error messages as follows:

```
Validation Error : 
Should be able to parse object at: spec/components/ at index: 0 
Please check for missing or incorrect elements

Validation Error : 
Should be able to parse object at: spec/components/ at index: 0 
Should be able to parse object at: spec/components/ at index: 1 
Please check for missing or incorrect elements
```

## Related Issues
#195 

## Release Information
next

## Consumer Impact
NA

## Testing
manually tested

### Unit Test(s)
Modified `test_manifest_validation` 

### Acceptance or Integration
NA

### Manual Verification
manually verified

### DIFF
--- a/src/server/handlers.rs
+++ b/src/server/handlers.rs
@@ -969,7 +969,7 @@ pub(crate) async fn validate_manifest(manifest: Manifest) -> anyhow::Result<()> 
             ));
         }
         return Err(anyhow!(
-            "Validation Error : \n{}Please check for missing or incorrect elements",
+            "Validation Error: \n{}Please check for missing or incorrect elements",
             error_message
         ));
     }
--- a/src/server/handlers.rs
+++ b/src/server/handlers.rs
@@ -3,7 +3,7 @@ use std::collections::{HashMap, HashSet};
 use anyhow::anyhow;
 use async_nats::{jetstream::stream::Stream, Client, Message};
 use base64::{engine::general_purpose::STANDARD as B64decoder, Engine};
-use jsonschema::{Draft, JSONSchema};
+use jsonschema::{paths::PathChunk, Draft, JSONSchema};
 use regex::Regex;
 use serde_json::json;
 use tokio::sync::OnceCell;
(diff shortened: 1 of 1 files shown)


src/server/handlers.rs
```diff
@@ -952,14 +952,26 @@ pub(crate) async fn validate_manifest(manifest: Manifest) -> anyhow::Result<()>
...
-                error.instance,
-                error.instance_path
+                "Should be able to parse object at: {} \n",
+                // The path of the corresponding JSON error instance in that file
+                instance_path
             ));
         }
-        return Err(anyhow!("Validation Error : \n{}", error_message));
+        return Err(anyhow!(
+            "Validation Error : \n{}Please check for missing or incorrect elements",
```
brooksmtownsend, line 972: nit: we can remove the space between `Error :`

brooksmtownsend (dismissed):
This makes the best of what we got! 😄 This should help just narrow down the source of the error

brooksmtownsend (approved):

PR #222, opened 2023-11-23
Derive PartialEq and Eq for DeleteResult

joonas:
## Feature or Problem

`DeleteResult` enum is missing `PartialEq` and `Eq`, which seem to be implemented for all of the other `<Operation>Result` types.


joonas:
Note, for the specific use case I have, I was able to work around this with `matches!(...)`, but it would be nice to be able to make use of the equality comparisons instead.

connorsmith256 (approved):

PR #223, opened 2023-11-28
feat(status)!: rename compensating and ready status

### LINKED ISSUE #177: Rename application statuses to more descriptive and used words
Currently, our [application statuses](https://github.com/wasmCloud/wadm/blob/main/src/server/types.rs#L231) are as follows:
```rust
/// All possible status types
#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, Clone, Copy, Default)]
#[serde(rename_all = "lowercase")]
pub enum StatusType {
    #[default]
    Undeployed,
    Compensating,
    Ready,
    Failed,
}
```

They cover a good variety of cases, but aren't necessarily the best naming conventions. The statuses don't inherently have anything wrong with them but ideally they reflect status types that orchestrators/control planes like [Nomad](https://www.nomadproject.io/) and [Kubernetes](https://kubernetes.io/) uses. I'd like to throw out a strawman and go from there 😄 

Here's what I would love to rename the statuses to:
1. Undeployed stays the same
2. Compensating -> Reconciling (To reflect when wadm has taken actions towards desired state and is awaiting events that would reach desired state)
3. Ready -> Deployed
4. Failed stays the same. Similar orchestrators reflect failed states as `pending` when there aren't enough applicable nodes, but I think our use here is fine.

The above assumes we like our current `put` and then `deploy` model, which I personally like because you can fearlessly put new versions of an application and then rollout at a future date.

### Additional Information
These are the available statuses and types in Nomad
- `Status`: The job's current state. It can have one of the following values:
  - `pending`: The job is currently waiting on scheduling.
  - `running`: The job has non-terminal allocations.
  - `dead`: All of the job's allocations and evaluations are terminal.
- `Type`: The type of job in terms of scheduling. It can have one of the following values:
  - `service`: Allocations are intended to remain alive.
  - `batch`: Allocations are intended to exit.
  - `system`: Each client gets an allocation.

Kubernetes follows a similar set of statuses for pods, and we don't have a notion of a "succeeded" state for wasmcloud apps.
Value | Description
-- | --
Pending | The Pod has been accepted by the Kubernetes cluster, but one or more of the containers has not been set up and made ready to run. This includes time a Pod spends waiting to be scheduled as well as the time spent downloading container images over the network.
Running | The Pod has been bound to a node, and all of the containers have been created. At least one container is still running, or is in the process of starting or restarting.
Succeeded | All containers in the Pod have terminated in success, and will not be restarted.
Failed | All containers in the Pod have terminated, and at least one container has terminated in failure. That is, the container either exited with non-zero status or was terminated by the system.
Unknown | For some reason the state of the Pod could not be obtained. This phase typically occurs due to an error in communicating with the node where the Pod should be running.

brooksmtownsend:
## Feature or Problem
This PR renames the `Compensating` status to `Reconciling`, and the `Ready` status to `Deployed`. According to discussion in #177 this is what most folks agreed was more intuitive than what we had, and Undeployed/Failed are still acceptable statuses.

## Related Issues
Fixes #177 

## Release Information
Next minor release

## Consumer Impact
Downstream consumers of the wadm crate will be able to simply replace all uses of compensating and ready to the new status enum member. Any consumers of the status stream that look for a specific status message will need to update to use the new status.


src/server/types.rs
```diff
@@ -231,8 +231,10 @@ impl StatusInfo {
 pub enum StatusType {
     #[default]
     Undeployed,
-    Compensating,
-    Ready,
+    #[serde(alias = "Compensating")]
```
connorsmith256, line 234: Does `alias` "stack" with `rename_all = "lowercase"`? I'm curious if this should be `compensating`?

connorsmith256 (approved):
LGTM! Just one question

src/server/types.rs
brooksmtownsend: Good question. Let me see if I can test that

connorsmith256 (approved):

PR #224, opened 2023-11-28
feat(status): prevent duplicate status publishes

### LINKED ISSUE #187: Make status stream less noisy
Right now, scalers will update the status stream consistently with a "ready" status even when nothing has changed. This can make the stream a bit noisy and push out useful data from previous statuses. We should update the logic to only push status changes when ready if the status changes from ready

brooksmtownsend:
## Feature or Problem
This PR adds a simple check before publishing a status to query the previous status and skip publishing if it would be the exact same. This should quiet down the status stream quite a bit for already deployed applications and it should make it a lot easier to follow during debugging.

## Related Issues
Fixes #187 

## Release Information
`next` release, can be a patch.

## Consumer Impact
This PR does add a burden on the status stream with a stream get

### Manual Verification
The status stream was much less noisy during e2e tests, for example during the upgrade e2e test the number of status messages went from 57 to 23. This is impressive as this PR isn't primarily concerned with reducing the number during deployment, we just end up publishing a lot of duplicate statuses.

### DIFF
--- a/src/workers/event_helpers.rs
+++ b/src/workers/event_helpers.rs
@@ -1,7 +1,8 @@
+use async_nats::jetstream::stream::Stream;
 use std::collections::{BTreeMap, HashMap};
 use std::fmt::Debug;
 
-use tracing::{instrument, warn};
+use tracing::{instrument, trace, warn};
 use wasmcloud_control_interface::{HostInventory, LinkDefinition};
 
 use crate::{commands::Command, publisher::Publisher, server::StatusInfo, APP_SPEC_ANNOTATION};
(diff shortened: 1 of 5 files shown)


src/workers/event_helpers.rs
```diff
@@ -116,12 +124,34 @@ impl<Pub> StatusPublisher<Pub> {
...
+        // worst case where the last message isn't all the way updated, we may publish a duplicate
+        // status. This is an acceptable tradeoff to not have to query the leader directly every time.
+        let prev_status = if let Some(status_stream) = &self.status_stream {
+            status_stream
+                .direct_get_last_for_subject(&topic)
+                .await
+                .map(|m| serde_json::from_slice::<StatusInfo>(&m.payload).ok())
+                .ok()
+                .flatten()
+        } else {
```
connorsmith256, line 139: Nit: how about `map` instead of `if let Some`, since you're not doing anything with the `None`?

connorsmith256 (approved):

src/workers/event_helpers.rs
brooksmtownsend: Oh I didn't do this because of the `await`, didn't want to introduce hackiness with an async closure

connorsmith256 (approved):

PR #231, opened 2024-01-08
chore: Replace atty with std::io::IsTerminal

joonas:
## Feature or Problem

As best as I can tell, [`atty` is unmaintained](https://github.com/softprops/atty/), which is unfortunate because it has an outstanding security vulnerability ([RUSTSEC-2021-0145](https://rustsec.org/advisories/RUSTSEC-2021-0145) / [GHSA-g98v-hv3f-hcfr](https://github.com/advisories/GHSA-g98v-hv3f-hcfr)).

Fortunately, [Rust 1.70.0 shipped native functionality to replace what `atty` was providing](https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html#isterminal).

I might've missed it, since I don't believe we have a defined msrv, I think expecting Rust 1.70.0 for this functionality seems fair.

If this is a bad assumption or we absolutely insist on supporting [Hermit](https://github.com/hermit-os), we should at least consider switching from `atty` to [`is-terminal`](https://crates.io/crates/is-terminal/), which is a drop-in replacement for `atty` but maintained.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,17 +177,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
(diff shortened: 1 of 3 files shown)


joonas:
not entirely sure about that failed test, is it known to be flaky or is it something I should look into?

thomastaylor312 (approved):
Yep, we've been doing this in other projects so it is about time we did it here. Thanks @joonas!

PR #233, opened 2024-01-10
feat(*)!: prevent querying inventory for new heartbeat

brooksmtownsend:
## Feature or Problem
This PR sets up the HostHeartbeat struct that we use in Wadm to be able to deserialize either the wasmCloud v0.81.0 version of the heartbeat or the v0.82.0 version of the heartbeat. Once v0.82.0 is released, we can remove the code branches handling the v81 heartbeat and take advantage of the additional information there to prevent querying host inventory on each heartbeat.

## Related Issues
Handles https://github.com/wasmCloud/wasmCloud/pull/1160
Fixes https://github.com/wasmCloud/wadm/issues/191

## Release Information
v0.10.0

## Consumer Impact
Backwards compatible with older versions of the host.

### Unit Test(s)
Some unit tests were modified to rip out old inventory format code. Additionally, tests were modified to support wash 0.25.0 and some new control interface command formats.

### Acceptance or Integration
All tests were updated to wash 0.25.0, wasmcloud 0.81.0, and NATS 2.10.7

PR #234, opened 2024-01-12
fix(scaler): always publish delete notification

brooksmtownsend:
## Feature or Problem
This PR fixes a possible issue where, in the case where a single wadm instance doesn't find scalers for a particular manifest name, it won't notify the other instances to delete those scalers as well. After this fix, anytime we request to remove scalers (which is only when a manifest is undeployed) we will send the `DeleteScalers` notification to each instance.

## Release Information
Can be a patch release cherry picked for 0.9.2, or just pushed into 0.10.0.


thomastaylor312 (approved):

thomastaylor312:
Good to go once tests pass

brooksmtownsend:
> Good to go once tests pass

Just a heads up that some of the test fixes went into #233, really just updating to v0.81 and the "new" heartbeat/inventory format

PR #238, opened 2024-01-16
feat(events): support actor_scaled event

brooksmtownsend:
## Feature or Problem
This PR adds a handler for the actor_scaled event which will be able to summarize all of the information in `actors_started` and `actors_stopped` with an idempotent event.

Though there's future work to really make this work with v0.82, this can be released at any time without problems as the event doesn't exist yet.

## Related Issues
Requires https://github.com/wasmCloud/wasmCloud/pull/1336
Filed #239 to cover future work that can't be done today.

## Release Information
v0.11.0

## Consumer Impact
Note that, if testing with the unreleased v0.82.0, this may cause duplicate state updates because both the actors_started and actor_scaled events will be handled. I'll be creating an issue shortly to cover this update.

### Unit Test(s)
I modified the `test_all_state` function to ensure functionality around starting/stopping/scaling up/down actors.

### DIFF
--- a/src/workers/event.rs
+++ b/src/workers/event.rs
@@ -208,6 +208,87 @@ where
         Ok(())
     }
 
+    #[instrument(level = "debug", skip(self, actor), fields(actor_id = %actor.public_key, host_id = %actor.host_id))]
+    async fn handle_actor_scaled(
+        &self,
+        lattice_id: &str,
+        actor: &ActorScaled,
+    ) -> anyhow::Result<()> {
+        trace!("Scaling actor in store");
+        debug!("Fetching current data for actor");
+
+        // Update actor count in the actor state, adding to the state if it didn't exist or removing
+        // if the scale is down to zero.
+        let mut actor_data = Actor::from(actor);
+        if let Some(mut current) = self
+            .store
+            .get::<Actor>(lattice_id, &actor.public_key)
+            .await?
+        {
+            trace!(actor = ?current, "Found existing actor data");
+
+            match current.instances.get_mut(&actor.host_id) {
+                // If the actor is running and is now scaled down to zero, remove it
+                Some(current_instances) if actor.max_instances == 0 => {
+                    current_instances.remove(&actor.annotations);
+                }
+                // If an actor is already running on a host, update the running count to the scaled max_instances value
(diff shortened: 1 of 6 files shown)


src/workers/event.rs
```diff
@@ -208,6 +208,87 @@ where
...
     }
 
+    #[instrument(level = "debug", skip(self, actor), fields(actor_id = %actor.public_key, host_id = %actor.host_id))]
+    async fn handle_actor_scaled(
+        &self,
+        lattice_id: &str,
+        actor: &ActorScaled,
+    ) -> anyhow::Result<()> {
+        trace!("Scaling actor in store");
+        debug!("Fetching current data for actor");
```
connorsmith256, line 218: ~Nit: are both of these logs needed?~ Nvm, I see we do this everywhere, just confused why

connorsmith256 (approved):

src/workers/event.rs
brooksmtownsend: Yeah good point, the back to back makes it a little bit weirder but it is following the other pieces of code where we debug log around fetching data.

thomastaylor312: I think that, once upon a time, there was stuff in between them. These ones could probably be combined, but I can see places where we might want to log something super deep at a trace level and then something else at a debug level

test/data/events.json
```diff
@@ -62,32 +62,6 @@
     "time": "2023-02-14T19:21:14.113113Z",
     "type": "com.wasmcloud.lattice.health_check_status"
   },
```
thomastaylor312, line 64: Can you please add a sample ActorScaled and ActorScaleFailed event in this file? Just want to make sure we catch any deserialization problems as we make changes

thomastaylor312 (changes requested):
All good to go other than one small blocking change

thomastaylor312 (approved):

brooksmtownsend:
Holding on the merge here just a moment to ensure we can release v0.82 before we start cutting wadm releases with this change. Only reason why is that here, for v0.82, we'll handle both the actorscaled and actorsstarted events so we might get incorrect state inbetween heartbeats.

PR #241, opened 2024-01-19
feat(*): subscribe to updated lattice event subjects

connorsmith256:
## Feature or Problem
The wasmCloud host now emits events on `wasmbus.evt.{lattice-id}.{event-type}`, so this updates wadm to use the suffixed subject

## Release Information
~Next. Not a breaking change, so maybe v0.10.1?~
v0.11.0

## Consumer Impact
N/A

## Testing
Updated existing tests


brooksmtownsend (commented):
Added in one more commit here after syncing offline to refactor the lattice parsing information and ensure we're listening on the wildcard event subject 👍🏻

brooksmtownsend (approved):

PR #243, opened 2024-01-30
feat: support TLS CA certificates when building a NATS client

protochron:
## Feature or Problem
NATS servers support TLS, which means sometimes you need to be able to provide certificates when instantiating a client.

This specifically adds support for supplying a CA certificate to use when connecting to a NATS server. I opted not to add client certificate support for mTLS for now, but it's an easy lift if it should just be added now.

## Testing
Tested locally using a NATS server with a self-signed CA and cert. Everything worked fine

### DIFF
--- a/bin/main.rs
+++ b/bin/main.rs
@@ -111,6 +111,10 @@ struct Args {
     )]
     nats_creds: Option<PathBuf>,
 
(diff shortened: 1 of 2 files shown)


brooksmtownsend (approved):
Nice!

PR #248, opened 2024-03-18
feat: Add helm chart for deploying wadm on Kubernetes

joonas:
## Feature or Problem

This adds a helm chart for deploying wadm on Kubernetes

### DIFF
--- a/charts/wadm/templates/deployment.yaml
+++ b/charts/wadm/templates/deployment.yaml
@@ -1,0 +1,116 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "wadm.fullname" . }}
+  labels:
+    {{- include "wadm.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "wadm.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "wadm.labels" . | nindent 8 }}
+        {{- with .Values.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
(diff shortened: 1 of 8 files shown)


charts/wadm/values.yaml
```diff
@@ -0,0 +1,66 @@
...
+    drop:
+      - "ALL"
+  seccompProfile:
+    type: "RuntimeDefault"
+
+
+env:
+  - name: WADM_NATS_SERVER
+    value: nats-cluster.default.svc.cluster.local
+  - name: WADM_JETSTREAM_DOMAIN
```
thomastaylor312, line 47: I wouldn't set anything as a default here. By default it doesn't use a domain in most setups and that is fine

charts/wadm/values.yaml
```diff
@@ -0,0 +1,66 @@
...
+  runAsNonRoot: true
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+      - "ALL"
+  seccompProfile:
+    type: "RuntimeDefault"
+
+
+env:
```
thomastaylor312, line 44: Can you switch this out to do something similar to the wasmcloud chart and also enumerate all of the configurable options? e.g.

```yaml
wadm:
    nats_server: nats-cluster.default.svc.cluster.local
    # Rest of the options here, commented out and documented
```

thomastaylor312 (changes requested):
Just one small change around how to expose configuration options

joonas:
@thomastaylor312 I believe I've got all of your feedback addressed, PTAL

charts/wadm/templates/deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+            - name: WADM_API_PREFIX
+              value: {{ .Values.wadm.config.apiPrefix }}
+          {{- end }}
+          {{- if .Values.wadm.config.streamPrefix }}
+            - name: WADM_STREAM_PREFIX
+              value: {{ .Values.wadm.config.streamPrefix }}
+          {{- end }}
+          {{- if .Values.wadm.config.multitenant }}
+            - name: WADM_MULTITENANT
+              value: {{ .Values.wadm.config.multitenant }}
```
thomastaylor312, line 101: These boolean values need to be quoted or they will be interpreted as a boolean rather than the string k8s expects. There are a few others here too. Other option might be to quote them in the values file

thomastaylor312 (changes requested):
Basically there, just one more small change

charts/wadm/templates/deployment.yaml
joonas: Oh that's right, I've gone ahed and added quote for each, thank you for the reminder!

thomastaylor312 (approved):

PR #254, opened 2024-04-04
feat(model): support labels in manifest

### LINKED ISSUE #249: [BUG] Manifests do not support labels
It looks like we never got around to supporting labels in the OAM manifest spec for Wadm. Just like Kubernetes, the metadata section of a manifest should support arbitrary key/value pairs. See https://github.com/oam-dev/spec/blob/e0991c5d0e666333ebbfa19f3490320bf671d8c0/metadata.md for what the spec calls for.

We may want to also enforce naming requirements for both [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) and [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) while we're at it.

brooksmtownsend:
## Feature or Problem
This PR adds supports for serializing the labels in a manifest and validating that they are correct according to the kubernetes label specification, which also matches the OAM specification.

## Related Issues
Fixes #249 

## Release Information
Next wadm

### Unit Test(s)
Added unit test to sanity check validation rules.

### Manual Verification
With a manifest:
```yaml
kind: Application
metadata:
  name: kvcounter-rust
  annotations:
    version: v0.0.1
    description: "Kvcounter demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)"
  labels:
    app.kubernetes.io/name: kvcounter-rust
spec:
```

The requested spec comes back as (including the label):
```bash
➜ nats req "wadm.api.default.model.get.kvcounter-rust" '{}'
17:15:21 Sending request on "wadm.api.default.model.get.kvcounter-rust"
17:15:21 Received with rtt 2.827708ms
{"result":"success","message":"Successfully fetched model kvcounter-rust","manifest":{"apiVersion":"core.oam.dev/v1beta1","kind":"Application","metadata":{"name":"kvcounter-rust","annotations":{"description":"Kvcounter demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)","version":"v0.0.1"},"labels":{"app.kubernetes.io/name":"kvcounter-rust"}}
```

### DIFF
--- a/src/server/handlers.rs
+++ b/src/server/handlers.rs
@@ -1,6 +1,6 @@
 use std::collections::{HashMap, HashSet};
 
-use anyhow::{anyhow, bail};
+use anyhow::{anyhow, bail, ensure};
 use async_nats::{jetstream::stream::Stream, Client, Message, Subject};
 use base64::{engine::general_purpose::STANDARD as B64decoder, Engine};
 use jsonschema::{paths::PathChunk, Draft, JSONSchema};
@@ -976,6 +976,9 @@ pub(crate) async fn validate_manifest(manifest: Manifest) -> anyhow::Result<()> 
         ));
     }
 
+    ensure!(manifest.metadata.labels.iter().all(valid_oam_label));
+    ensure!(manifest.metadata.annotations.iter().all(valid_oam_label));
+
     for component in manifest.spec.components.iter() {
         // Component name validation : each component (actors or providers) should have a unique name
         if !name_registry.insert(component.name.clone()) {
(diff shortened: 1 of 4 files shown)


src/server/handlers.rs
```diff
@@ -1198,4 +1256,28 @@ mod test {
...
+        // Valid labels
+        assert!(valid_oam_label((&"foo".to_string(), &"bar".to_string())));
+        assert!(valid_oam_label((
+            &"app.oam.io/name".to_string(),
+            &"wasmcloud".to_string()
+        )));
+        assert!(valid_oam_label((
+            &"app.oam.io/name".to_string(),
+            &"wasmcloud".to_string()
+        )));
```
joonas, line 1271: Are these supposed to be the same?

brooksmtownsend: Just to be double sure...

Nope, just be a copy paste fail. Will change one

joonas (approved):
Looks good to me!

src/server/handlers.rs
```diff
@@ -1075,6 +1078,61 @@ fn parse_image_ref(image_name: &str) -> Option<(String, String)> {
...
+fn is_valid_dns_subdomain(s: &str) -> bool {
+    if s.is_empty() || s.len() > 253 {
+        return false;
+    }
+
+    s.split('.').all(|part| {
+        !part.is_empty()
+            && part.len() <= 63
+            && part.chars().all(|c| c.is_ascii_alphanumeric() || c == '-')
+            && part
```
thomastaylor312, line 1103: Are these last two checking for leading and trailing whitespace? Would it just be clearer to use `starts_with` and `ends_with` instead?

src/server/handlers.rs
```diff
@@ -1075,6 +1078,61 @@ fn parse_image_ref(image_name: &str) -> Option<(String, String)> {
...
+/// This function validates both a name and a value with the same rules.
+fn is_valid_label_value(name: &str) -> bool {
+    if name.is_empty() || name.len() > 63 {
+        return false;
+    }
+
+    name.chars()
+        .all(|c| c.is_ascii_alphanumeric() || c == '-' || c == '_' || c == '.')
+        && name
+            .chars()
```
thomastaylor312, line 1125: Same here around whitespace

src/server/handlers.rs
```diff
@@ -1198,4 +1256,28 @@ mod test {
...
+        assert!(valid_oam_label((
+            &"app.oam.io/name".to_string(),
+            &"wasmcloud".to_string()
+        )));
+        assert!(valid_oam_label((
+            &"justaregularstring".to_string(),
+            &"wasmcloud".to_string()
+        )));
+
+        // Invalid labels
```
thomastaylor312, line 1273: Do we need some checking for whitespace?

thomastaylor312 (changes requested):
One blocking comment around things to test for. Others are optional

src/server/handlers.rs
brooksmtownsend: It's actually checking to see if the first and last character is an alphanumeric character

src/server/handlers.rs
brooksmtownsend: Yeah will be good to add a test for this

src/server/handlers.rs
brooksmtownsend: Will check out using the starts_with with a pattern to make this better. Also I can leave a comment here to say why I'm checking

src/server/handlers.rs
```diff
@@ -1075,6 +1078,50 @@ fn parse_image_ref(image_name: &str) -> Option<(String, String)> {
...
+    if s.is_empty() || s.len() > 253 {
+        return false;
+    }
+
+    s.split('.').all(|part| {
+        // Ensure each part is non-empty, <= 63 characters, starts with an alphabetic character,
+        // ends with an alphanumeric character, and contains only alphanumeric characters or hyphens
+        !part.is_empty()
+            && part.len() <= 63
+            && part.starts_with(|c: char| c.is_ascii_alphabetic())
```
thomastaylor312, line 1104: Super nit: you could just do `part.starts_with(char::is_ascii_alphabetic)`

thomastaylor312 (approved):

PR #272, opened 2024-05-01
Add support for mounting Kubernetes Secret for nats.creds file

joonas:
## Feature or Problem

We want the ability to either directly reference NATS jwt / seed, or have the ability to pass them in via Kubernetes Secret that gets volume mounted into the pod. This adds support for the volume mounting the secret.


protochron (approved):

charts/wadm/values.yaml
```diff
@@ -14,12 +14,14 @@ wadm:
...
-      credsFile: ""
-      jetstreamDomain: ""
-      jwt: ""
-      seed: ""
       server: "127.0.0.1:4222"
+      jetstreamDomain: ""
       tlsCaFile: ""
+      creds:
+        jwt:
+        seed:
```
brooksmtownsend, line 22: Are these keys of no value equivalent to specifying `""`

brooksmtownsend (commented):
Just a question to make sure leaving JWT+Seed blank is fine, otherwise LGTM 👍🏻

charts/wadm/values.yaml
joonas: That's correct, and honestly there's no reason to have the double quotes missing here, so I went ahead and added those back for these values as well.

brooksmtownsend (dismissed):
Test failed now? Weird. Good to go once resolved!

brooksmtownsend (approved):

PR #274, opened 2024-05-06
feat(validation): add manifest validation

vados-cosmonic:
## Feature or Problem

This commit adds validation functions along with utility accessors to `Manifest` to enable checking WADM manifests for common errors.

As the validation functions are exposed they can be used by downstream crates (ex. `wash`) to validate WADM manifests or try to catch errors *before* a manifest is used.

This is up for review in draft state so we can talk about the design -- if it looks good, I'll do the following:
- [x] Finish adding all the known interfaces

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/2009

## Release Information

`next`

### Unit Test(s)

Unit tests with fixtures have been added

### DIFF
--- a/src/model/mod.rs
+++ b/src/model/mod.rs
@@ -3,6 +3,7 @@ use std::collections::{BTreeMap, HashMap};
 use serde::{Deserialize, Serialize};
 
 pub(crate) mod internal;
+pub mod validation;
 
 /// The default weight for a spread
 pub const DEFAULT_SPREAD_WEIGHT: usize = 100;
@@ -58,6 +59,38 @@ impl Manifest {
             .annotations
             .get(DESCRIPTION_ANNOTATION_KEY)
             .map(|v| v.as_str())
+    }
+
+    /// Returns the components in the manifest
+    pub fn components(&self) -> impl Iterator<Item = &Component> {
+        self.spec.components.iter()
+    }
+
+    /// Returns only the WebAssembly components in the manifest
+    pub fn wasm_components(&self) -> impl Iterator<Item = &Component> {
+        self.components()
+            .filter(|c| matches!(c.properties, Properties::Component { .. }))
+    }
+
+    /// Returns only the provider components in the manifest
+    pub fn capability_providers(&self) -> impl Iterator<Item = &Component> {
+        self.components()
+            .filter(|c| matches!(c.properties, Properties::Capability { .. }))
+    }
+
+    /// Returns a map of component names to components in the manifest
+    pub fn component_lookup(&self) -> HashMap<&String, &Component> {
+        self.components()
+            .map(|c| (&c.name, c))
+            .collect::<HashMap<&String, &Component>>()
+    }
+
+    /// Returns only links in the manifest
+    pub fn links(&self) -> impl Iterator<Item = &Trait> {
+        self.components()
+            .flat_map(|c| c.traits.as_ref())
+            .flatten()
+            .filter(|t| t.is_link())
     }
 }
 
@@ -151,6 +184,11 @@ impl Trait {
             trait_type: LINK_TRAIT.to_owned(),
             properties: TraitProperty::Link(props),
         }
+    }
+
+    /// Check if a trait is a link
+    pub fn is_link(&self) -> bool {
+        self.trait_type == LINK_TRAIT
     }
 
     /// Helper that creates a new spreadscaler type trait with the given properties
--- a/src/model/validation.rs
+++ b/src/model/validation.rs
@@ -1,0 +1,435 @@
+//! Logic for model ([`Manifest`]) validation
+//!
+
+use std::collections::HashMap;
+use std::path::Path;
+use std::sync::OnceLock;
+
+use anyhow::{Context as _, Result};
+use regex::Regex;
+use serde::{Deserialize, Serialize};
+
+use crate::model::{LinkProperty, Manifest, TraitProperty, LATEST_VERSION};
+
+/// A namespace -> package -> interface lookup
+type KnownInterfaceLookup = HashMap<String, HashMap<String, HashMap<String, ()>>>;
+
+/// Hard-coded list of known namespaces/packages and the interfaces they contain.
+///
+/// Using an interface that is *not* on this list is not an error --
+/// custom interfaces are expected to not be on this list, but when using
+/// a known namespace and package, interfaces should generally be well known.
+static KNOWN_INTERFACE_LOOKUP: OnceLock<KnownInterfaceLookup> = OnceLock::new();
+
+/// Get the static list of known interfaces
+fn get_known_interface_lookup() -> &'static KnownInterfaceLookup {
+    KNOWN_INTERFACE_LOOKUP.get_or_init(|| {
+        HashMap::from([
+            (
+                "wrpc".into(),
+                HashMap::from([
+                    (
+                        "blobstore".into(),
+                        HashMap::from([("blobstore".into(), ())]),
+                    ),
+                    (
+                        "keyvalue".into(),
+                        HashMap::from([("atomics".into(), ()), ("store".into(), ())]),
+                    ),
+                    (
+                        "http".into(),
+                        HashMap::from([
+                            ("incoming-handler".into(), ()),
+                            ("outgoing-handler".into(), ()),
+                        ]),
+                    ),
+                ]),
+            ),
+            (
+                "wasi".into(),
+                HashMap::from([
+                    (
+                        "blobstore".into(),
+                        HashMap::from([("blobstore".into(), ())]),
+                    ),
+                    ("config".into(), HashMap::from([("runtime".into(), ())])),
+                    (
+                        "keyvalue".into(),
+                        HashMap::from([("atomics".into(), ()), ("store".into(), ())]),
+                    ),
+                    (
+                        "http".into(),
+                        HashMap::from([
+                            ("incoming-handler".into(), ()),
+                            ("outgoing-handler".into(), ()),
+                        ]),
+                    ),
+                    ("logging".into(), HashMap::from([("logging".into(), ())])),
+                ]),
+            ),
+            (
+                "wasmcloud".into(),
+                HashMap::from([(
+                    "messaging".into(),
+                    HashMap::from([("consumer".into(), ()), ("handler".into(), ())]),
+                )]),
+            ),
+        ])
+    })
+}
+
+static MANIFEST_NAME_REGEX_STR: &str = r"^[-\w]+$";
+static MANIFEST_NAME_REGEX: OnceLock<Regex> = OnceLock::new();
+
+/// Retrieve regular expression which manifest names must match, compiled to a usable [`Regex`]
+fn get_manifest_name_regex() -> &'static Regex {
+    MANIFEST_NAME_REGEX.get_or_init(|| {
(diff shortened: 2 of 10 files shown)


src/model/validation.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// Get the static list of known interfaces
+fn get_known_interface_lookup() -> &'static KnownInterfaceLookup {
+    KNOWN_INTERFACE_LOOKUP.get_or_init(|| {
+        HashMap::from([
+            (
+                "wrpc".into(),
+                HashMap::from([
+                    ("keyvalue".into(), HashMap::from([("atomics".into(), ())])),
+                    ("blobstore".into(), HashMap::from([("atomics".into(), ())])),
```
brooksmtownsend, line 36: blobstore should be `blobstore` not `atomics` for the interfaces right?

src/model/validation.rs
```diff
@@ -0,0 +1,242 @@
...
+                    ("config".into(), HashMap::from([("runtime".into(), ())])),
+                ]),
+            ),
+            (
+                "wasi".into(),
+                HashMap::from([
+                    ("keyvalue".into(), HashMap::from([("atomics".into(), ())])),
+                    ("blobstore".into(), HashMap::from([("atomics".into(), ())])),
```
brooksmtownsend, line 44: Same comment re: blobstore/atomics

src/model/mod.rs
```diff
@@ -28,6 +29,9 @@ pub const LINK_TRAIT: &str = "link";
 /// for a manifest
 pub const LATEST_VERSION: &str = "latest";
 
+/// Type Alias for names of components
+type ComponentName = String;
+
```
brooksmtownsend, line 34: I would prefer just using a `String` in the functions below, especially since it's a public API, and using doc comments to indicate that its a component name

src/model/validation.rs
```diff
@@ -0,0 +1,242 @@
...
+use std::sync::OnceLock;
+
+use anyhow::{Context as _, Result};
+use serde::{Deserialize, Serialize};
+
+use crate::model::{LinkProperty, Manifest, TraitProperty};
+
+type NamespaceName = String;
+type PackageName = String;
+type InterfaceName = String;
```
brooksmtownsend, line 15: Same comments re: simple aliases since this is a public API

brooksmtownsend (commented):
Looks real nice so far, just wanted to double check on the interfaces (since I saw your comment about adding the rest) and around type aliases

src/model/validation.rs
vados-cosmonic: See [this comment](https://github.com/wasmCloud/wadm/pull/274#discussion_r1591390096) -- the list in the PR is nowhere near complete but I was more looking for any design feedback!

src/model/mod.rs
vados-cosmonic: Sure!

- [ ] Remove `type` aliases for `String`s

vados-cosmonic:
Some more stuff after discussion with @brooksmtownsend that we should probably add to this:

- [ ] API that handles strings for easy validation (probably a small refactor to use this from file-path based one)
- [ ] Merge checks from `server/handler.rs`

src/model/validation.rs
```diff
@@ -0,0 +1,424 @@
...
+                    ("logging".into(), HashMap::from([("logging".into(), ())])),
+                ]),
+            ),
+            (
+                "wasmcloud".into(),
+                HashMap::from([
+                    (
+                        "bus".into(),
+                        HashMap::from([("lattice".into(), ()), ("host".into(), ())]),
+                    ),
```
brooksmtownsend, line 71: This one should just be a builtin, so we shouldn't specify a link in the manifest right?

src/model/validation.rs
```diff
@@ -0,0 +1,424 @@
...
+fn is_invalid_known_interface(namespace: &str, package: &str, interface: &str) -> bool {
+    let known_interfaces = get_known_interface_lookup();
+    let Some(pkg_lookup) = known_interfaces.get(namespace) else {
+        // This namespace isn't known, so it may be a custom interface
+        return false;
+    };
+    let Some(iface_lookup) = pkg_lookup.get(package) else {
+        // Unknown package inside a known interface we control is probably a bug
+        return true;
+    };
```
brooksmtownsend, line 143: I'm a _little_ concerned with this one when it comes to wadm supporting interfaces in a forwards compatible way, since this would also necessitate a wadm upgrade when, say, `wasi:keyvalue` adds a new interface 🤔 

I totally get what this is supposed to prevent though, linking on like `wasi:http/incoming-handlerr` and being frustrated endlessly. Do you have thoughts on the tradeoff here?

brooksmtownsend (changes requested):
A nit or two, a couple requests and a discussion point that I might just be overthinking. This looks great and I love the reusability of the library!

src/model/validation.rs
```diff
@@ -0,0 +1,424 @@
...
+    KNOWN_INTERFACE_LOOKUP.get_or_init(|| {
+        HashMap::from([
+            (
+                "wrpc".into(),
+                HashMap::from([
+                    (
+                        "blobstore".into(),
+                        HashMap::from([("blobstore".into(), ())]),
+                    ),
+                    ("config".into(), HashMap::from([("runtime".into(), ())])),
```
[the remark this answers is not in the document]
vados-cosmonic: Sure, will remove!

src/model/validation.rs
```diff
@@ -0,0 +1,424 @@
...
+    })
+}
+
+/// Check whether a manifest name matches requirements, returning all validation errors
+pub fn validate_manifest_name(name: &str) -> impl ValidationOutput {
+    let mut errors = Vec::new();
+    if !get_manifest_name_regex().is_match(name) {
+        errors.push(ValidationFailure::new(
+            ValidationFailureLevel::Error,
+            format!("manifest name [{name}] is not allowed (shoudl match regex [{MANIFEST_NAME_REGEX_STR}])"),
```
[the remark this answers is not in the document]
vados-cosmonic: Nice catch, fixed!

src/model/validation.rs
```diff
@@ -0,0 +1,424 @@
...
+                    ),
+                    ("config".into(), HashMap::from([("runtime".into(), ())])),
+                    (
+                        "keyvalue".into(),
+                        HashMap::from([("atomics".into(), ()), ("store".into(), ())]),
+                    ),
+                    (
+                        "http".into(),
+                        HashMap::from([("incoming-handler".into(), ())]),
+                    ),
```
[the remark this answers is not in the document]
vados-cosmonic: Added!

src/model/validation.rs
```diff
@@ -0,0 +1,424 @@
...
+                        HashMap::from([("blobstore".into(), ())]),
+                    ),
+                    ("config".into(), HashMap::from([("runtime".into(), ())])),
+                    (
+                        "keyvalue".into(),
+                        HashMap::from([("atomics".into(), ()), ("store".into(), ())]),
+                    ),
+                    (
+                        "http".into(),
+                        HashMap::from([("incoming-handler".into(), ())]),
```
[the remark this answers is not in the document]
vados-cosmonic: Added!

src/model/validation.rs
vados-cosmonic: Ah true -- we haven't used this in a dynamic way at all so far, will remove

src/model/validation.rs
vados-cosmonic: What I did for now was to make it a warning -- it can be up to downstream tools on whether to be strict or not

vados-cosmonic: What I did for now was to make it a warning -- it can be up to downstream tools on whether to be strict or not, what do you think @brooksmtownsend

src/model/validation.rs
```diff
@@ -0,0 +1,427 @@
...
+///
+/// ```rust,ignore
+/// let messages = validate_manifest(some_path).await?;
+/// let valid = messages.valid();
+/// ```
+///
+/// # Arguments
+///
+/// * `content` - YAML content  to the Manifest that will be read into memory and validated
+pub async fn validate_manifest(
```
thomastaylor312, line 299: I think we should probably rename this one to `validate_manifest_raw` and have it take an `impl AsRef<[u8]>` and then have a `validate_manifest` function that takes a `&Manifest` parameter for the cases when a manifest has already been parsed

src/server/handlers.rs
```diff
@@ -54,22 +54,15 @@ impl<P: Publisher> Handler<P> {
...
 
-        if manifest.version() == LATEST_VERSION {
-            self.send_error(
-                msg.reply,
-                format!("A manifest with a version {LATEST_VERSION} is not allowed in wadm"),
-            )
-            .await;
+        let manifest_validation_output = validate_manifest_version(manifest.version());
+        if let Some(first_err) = manifest_validation_output.errors().first() {
+            self.send_error(msg.reply, format!("invalid manifest version: {first_err}"))
```
thomastaylor312, line 59: I think we should probably return a comma delimited list of all errors so the user doesn't have to fix one, then find out about the next one

thomastaylor312 (changes requested):
Excellent job here. Just a few minor blocking things around the publicly exposed crate API and how we return errors to a user

src/model/validation.rs
brooksmtownsend: @vados-cosmonic I think that makes sense! So here a future inclusion of `wasi:nosql` for example would just be a "heads up, we didn't know about this one", but adding `wasi:keyvalue/eventual` (the old interface) would fail validation and catch it. LGTM

brooksmtownsend (commented):
My comments all addressed 👍🏻

brooksmtownsend (approved):
My comments all addressed 👍🏻

src/model/validation.rs
vados-cosmonic: Made this change, thanks! I called it `validate_manifest_bytes` though

src/server/handlers.rs
vados-cosmonic: Sure! I don't think the `send_error` machinery is really built for this (and this really isn't the right place to be catching failures anyway) but made this change!

src/server/handlers.rs
```diff
@@ -54,22 +54,26 @@ impl<P: Publisher> Handler<P> {
...
+        let manifest_validation_output = validate_manifest_version(manifest.version());
+        let manifest_validation_errors = manifest_validation_output.errors();
+        if !manifest_validation_errors.is_empty() {
             self.send_error(
                 msg.reply,
-                format!("A manifest with a version {LATEST_VERSION} is not allowed in wadm"),
+                format!(
+                    "invalid manifest version, errors: {:#?}",
+                    manifest_validation_errors
+                        .iter()
```
thomastaylor312, line 65: Nit: This could be into_iter and avoid the clone

thomastaylor312 (approved):

thomastaylor312:
Leaving it for @brooksmtownsend to merge since he was first reviewer

src/server/handlers.rs
thomastaylor312: Actually just realized we’re dealing with borrowed data at this point

brooksmtownsend (approved):

PR #281, opened 2024-05-14
feat(*)!: Makes version optional

thomastaylor312:
This PR makes two major changes. The first, and most important, is that versions are now optional. If no version field is passed, wadm will automatically generate a ULID to use as its version. This means that listing versions can be done in order by sorting them lexographically.

Second is an inversion of the delete behavior. Most of the time the delete endpoint in practice has been used to delete the whole application and not just a specific version. Now, by default, if you call the delete endpoint, it will delete all versions. You can still pass a now optional version to delete a single specific version.

While I was in the handlers, I also removed an unused field for orphaning undeploys

### DIFF
--- a/src/server/handlers.rs
+++ b/src/server/handlers.rs
@@ -312,10 +312,6 @@ impl<P: Publisher> Handler<P> {
             .await
     }
 
-    // NOTE(thomastaylor312): This is different than wadm 0.3. I found it remarkably confusing that
-    // you could delete something without undeploying it. So the new behavior is that if a manifest
-    // that is deployed is deleted, it is automatically undeployed, and we indicate that to the
-    // user. This should be documented when we get to our documentation tasks
     #[instrument(level = "debug", skip(self, msg))]
     pub async fn delete_model(
         &self,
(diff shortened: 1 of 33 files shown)


src/server/handlers.rs
```diff
@@ -312,10 +312,6 @@ impl<P: Publisher> Handler<P> {
             .await
     }
 
-    // NOTE(thomastaylor312): This is different than wadm 0.3. I found it remarkably confusing that
```
brooksmtownsend, line 315: tenor-606073

brooksmtownsend (changes requested):
Fully in support of making version optional, though I would request changing the commit to `feat(*)!` because not providing a version now is destructive instead of an error. On the `wash-lib` / `wash-cli` side we can still handle the `--delete-all` flag gracefully, and perhaps provide a warning, but it won't break existing functionality since not providing a version previously is an error

thomastaylor312:
@brooksmtownsend I updated the PR and the commit message on the squash merge to have the `!`

brooksmtownsend (approved):

PR #287, opened 2024-05-21
feat(*)!: Breaks wadm into separate crates and creates a new wadm-client

thomastaylor312:
This is a fairly large PR that splits out the wadm types into their own crate and also adds a new client library for interacting with the wadm API. Please see the individual commit messages for additional details


crates/wadm-types/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "wadm-types"
+description = "Types and validators for the wadm API"
+version = "0.11.2"
```
thomastaylor312, line 4: It felt like these should probably be the same version as current wadm to start, but I could easily be swayed for this to be 0.1.0 as well

crates/wadm-client/src/nats.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/// Takes a string that could be a raw seed, or a path and does all the necessary loading and parsing steps
+async fn get_seed(seed: String) -> Result<nkeys::KeyPair> {
+    // MAGIC NUMBER: Length of a seed key
+    let raw_seed = if seed.len() == 58 && seed.starts_with('S') {
+        seed
+    } else {
+        tokio::fs::read_to_string(seed).await?
+    };
```
brooksmtownsend, line 52: nit: it would be worth including a `.context` here to add the seed that we're assuming is a path, so that the error isn't just "no such file or directory"

crates/wadm-types/Cargo.toml
brooksmtownsend: I think 0.1.0 makes good sense, we may keep them aligned for a few releases but they'll quickly go out of sync

crates/wadm/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "wadm"
+description = "wasmCloud Application Deployment Manager: A tool for running Wasm applications in wasmCloud"
+version = "0.11.2"
```
brooksmtownsend, line 4: note for followup: will need to bump to 0.12.0 to release all this good stuff

brooksmtownsend (approved):
What a nice refactor 😍 I do like the idea of both of the new crates coming in at 1.0, and I left two quick comments that are optional

brooksmtownsend (approved):

PR #295, opened 2024-06-01
ref(*)!: Removes custom mirror code

thomastaylor312:
This removes our custom mirror code and replaces it with multiple filter subjects. While I was in here, I also changed our other `wadm.evt` topics to publish their event type in the topic as well. This is marked as a breaking change because it removes something from the crate API. This shouldn't affect normal users


brooksmtownsend (commented):
This looks good to me code-wise, so now wadm events and wasmCloud events are handled in the same place which is nice.

One thing I'd really like to make sure works before we merge this is that the upgrade process is seamless, trying this out now. Just want to make sure there aren't conflicts with the consumer name as the subject change

brooksmtownsend (changes requested):
@thomastaylor312 here's what I get trying to run this version of wadm after just running v0.12.0:
```
➜ cargo run --bin wadm
   Compiling wadm v0.12.0 (/Users/brooks/github.com/wasmcloud/wadm/crates/wadm)
   Compiling wadm-cli v0.12.0 (/Users/brooks/github.com/wasmcloud/wadm)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.99s
     Running `target/debug/wadm`
RUST_LOG was not set or the given directive was invalid: FromEnvError { kind: Env(NotPresent) }
Defaulting logger to `info` level
2024-06-07T13:06:07.843299Z  INFO async_nats::options: event: connected
2024-06-07T13:06:07.846296Z  WARN wadm::nats: Found stream wadm_events with different configuration, deleting and recreating
Error: Error { kind: JetStream(Error { code: 400, err_code: ErrorCode(10065), description: Some("subjects overlap with an existing stream") }), source: None }
```

To avoid a painful wash upgrade, can we have a check to delete the consumer (or, perhaps, to create a different one?) so that this doesn't bail?

thomastaylor312:
Ah I thought the delete did take care of that. Let me give that a whirl and fix it

thomastaylor312:
This was superseded by #310

Issue #296, opened 2024-06-03
[BUG] Wadm does not allow me to use the same provider with another component.


ritesh089:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

wash does not allow me to deploy the same provider with a second component that has its own wadm.
returns the error :
```
Provider file:///Users/****/Documents/wasm-dependencies/http-server-provider.par.gz is already deployed with a different version in http-component.
```
wash app status for the second component is undeployed .
```
Name               Latest Version   Deployed Version   Deploy Status   Description
  http-component     v0.0.1           v0.0.1                  Deployed   HTTP hello world demo in Rust, showing use of the server and client providers
  rust-hello-world   v0.0.1           N/A                   Undeployed   HTTP hello world demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)
  ```
  
  Note : I did use a different port for the second component , but the same provider.

### Steps to reproduce

1. wash up --wadm-version 0.11.2
2. wash app deploy wadm.yaml (for component1)

3. build component 2 : wash build
4. wash app deploy wadm.yaml (for component 2)

Wadm file : component 1
```
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: http-component
  annotations:
    version: v0.0.1
    description: 'HTTP hello world demo in Rust, showing use of the server and client providers'
spec:
  components:
    - name: http-component
      type: component
      properties:
        image: file://./build/http_component_s.wasm
      traits:
        # Govern the spread/scheduling of the component
        - type: spreadscaler
          properties:
            replicas: 1
        - type: link
          properties:
            target: httpclient
            namespace: wasi
            package: http
            interfaces: [outgoing-handler]

    # Add a capability provider that enables HTTP access
    - name: httpserver
      type: capability
      properties:
        image: file:///Users/****/Documents/wasm-dependencies/http-server-provider.par.gz
      traits:
        # Link the httpserver to the component, and configure the HTTP server
        # to listen on port 8080 for incoming requests
        - type: link
          properties:
            target: http-component
            namespace: wasi
            package: http
            interfaces: [incoming-handler]
            source_config:
              - name: default-http
                properties:
[...]

PR #297, opened 2024-06-03
feat(reuseProviders): Allow capability providers to be reused across components

ritesh089:
## Feature or Problem
Update wasmcloud deployment manager to allow reuse of capability providers across deployments.

## Related Issues
(https://github.com/wasmCloud/wadm/issues/296)

### DIFF
--- a/crates/wadm/src/server/handlers.rs
+++ b/crates/wadm/src/server/handlers.rs
@@ -600,37 +600,6 @@ impl<P: Publisher> Handler<P> {
             }
         }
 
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
Looks good! @ritesh089 and I talked offline, so just to bring that context here we decided this block was unnecessary since providers get their own assigned identifier and can be run on the same host multiple times.

Just to be clear, this does not allow for one provider instance to be used in multiple apps, this allows for one provider to have multiple instances in multiple apps.

PR #306, opened 2024-06-17
feat: add policy configuration block

protochron:
## Feature or Problem
This adds a policy block to the top-level configuration of the OAM specification. This block, defined in the [draft
specification](https://github.com/oam-dev/spec/blob/master/SPEC_DRAFT.md), is intended to be used for _application_ level configuration that applies to all components.

Initial use cases for this type of configuration include:
* configuring parameters to secret backends
* configuring Kubernetes Services, Ingresses, etc. configured by the wasmcloud-operator

An example of this might look something like:

```yaml
---
kind: Application
metadata:
  name: sample
  annotations:
    version: v0.0.1
    description: Sample manifest that passes
spec:
  policies:
    - name: aws-secret-config
      type: secret-backend
      properties:
        iam-role: some-role-id
    - name: k8s-service
      type: kubernetes-service
      properties:
        type: LoadBalancer
        # ...
  components:
    - name: http-component
      type: component
      properties:
        image: ghcr.io/wasmcloud/component-http-hello-world:0.1.0
      traits:
        - type: spreadscaler
          properties:
            replicas: 1
```

Instead of adding this new block to the existing JSON schema, this instead implements an OpenAPI v3 schema derived from the structs defined in the types crate and uses that to drive validation. This feels like much more maintainable approach going forward.


crates/wadm/src/server/handlers.rs
```diff
@@ -23,10 +21,6 @@ use crate::{model::StoredManifest, publisher::Publisher};
 
 use super::{parser::parse_manifest, storage::ModelStorage, ManifestNotifier};
 
-const JSON_SCHEMA: &str = include_str!("../../oam.schema.json");
```
thomastaylor312, line 26: Since you're removing this, can you also remove this file and the GH Actions check that makes sure this is in sync with the top level schema?

thomastaylor312 (changes requested):
Love this. Just one small ask around removing the schema file from the types library. If there is a way to generate a schema for people to consume, that would be nice as well since I know we have community members using the schema

crates/wadm/src/server/handlers.rs
brooksmtownsend: Other than the nondescript errors here, is there a good reason why we should not have a JSON schema embedded? I'm okay with this as long as the validation is done client-side to reasonably avoid errors, but I do like the assertion that an incoming manifest is valid

protochron: So the utoipa crate is actually validating that the manifest is correct, has all the right fields, etc.

Having not thought about this for ~2 weeks, I'm wondering if it actually may make more sense to revert back to using a jsonschema, but instead of using a static file, actually generate the schema from our model using [schemars](https://github.com/GREsau/schemars)

protochron:
@thomastaylor312 I added automatic jsonschema generation derived from the types directly and updated the CI check to ensure that any updates are committed. We should probably stick with OpenAPIv3 for the validation though since I think we'll want to move towards using that to drive the schema generation once the crate we're using supports 3.1

brooksmtownsend (approved):

thomastaylor312 (approved):

PR #307, opened 2024-06-17
feat!: add support for secrets in manifests

protochron:
## Feature or Problem
This adds support for secrets in wasmCloud application manifests. The
secrets themselves are actually _secret references_ as outlined in
wasmCloud/wasmCloud#2190. Just like config, secrets can be specified at
the component or provider level or on a link.

Secret references themselves are actually implemented as an additional
kind of config stored in the same config data bucket. However, I opted
to implement a dedicated scaler for secrets that is largely a clone of
the existing ConfigScaler since the underlying data type is very
different than the arbitrary set of key/value pairs we use for config.

An example of what this looks like in a component is shown below:

```yaml
spec:
  components:
    - name: http-component
      type: component
      properties:
        image: ghcr.io/wasmcloud/test-fetch-with-token:0.1.0-fake
        secrets:
          - name: some-api-token
            source:
              backend: nats-kv
              key: test-value
              version: 1
          - name: my-other-secret
            source:
              backend: aws-secrets-manager
              value: secret-name
              version: "be01a5fb-7ebb-4ae9-8ea0-0902e8940bc0"
```

This contains a breaking change to the way that we specify config on
links:

```yaml
- type: link
  properties:
    namespace: wasmcloud
    package: postgres
    interfaces: [managed-query]
    target:
      name: sql-postgres
      secrets:
        - name: db-password
          source:
            backend: nats-kv
            key: myapp_db-password
            version: 1
```

Instead of using `target_config` and `source_config`, this renames them
to `target` and `source` respectively and adds keys for `config` and
`secrets`. The name of the target is now a key at the top
level of the `target` block, as seen above.


crates/wadm/src/scaler/secretscaler.rs
```diff
@@ -0,0 +1,119 @@
...
+            secret_source,
+            status: RwLock::new(StatusInfo::reconciling("")),
+        }
+    }
+}
+
+#[async_trait]
+impl<S: SecretSource + Send + Sync + Clone> Scaler for SecretScaler<S> {
+    fn id(&self) -> &str {
+        todo!()
```
brooksmtownsend, line 48: :scream:

since this is a uniqueness check, combination of secret name + source should be enough?

crates/wadm/src/scaler/manager.rs
```diff
@@ -880,6 +900,15 @@ pub(crate) fn compute_component_id(
     }
 }
 
+pub(crate) fn compute_secret_id(
+    model_name: &str,
+    component_id: Option<&String>,
+    component_name: &str,
+) -> String {
+    let name = compute_component_id(model_name, component_id, component_name);
+    format!("secret_{}", name)
```
brooksmtownsend, line 909: Any thoughts on if this should be `secret_` or `SECRET_`? The uppercase matches our current bucket naming strategy but I'm not sure that we specifically need that

protochron: I don't have a strong preference here. Uppercase works if we want to stick with that convention

crates/wadm/src/scaler/secretscaler.rs
protochron: This was just missing the implementation of returning `&self.id`. Just something I didn't get a chance to fill in! The id uses the same hashing logic that the configscaler uses

crates/wadm/src/scaler/manager.rs
protochron: @brooksmtownsend I updated it to be `SECRET_` since uppercase made it seem much more like a generated value.

brooksmtownsend: Note to myself that I should change the PR in wasmCloud to use the uppercase convention

brooksmtownsend (changes requested):
One or two nits, and a bigger request around backwards compatible manifests. I found myself wondering if we needed a SecretScaler at all given that it's just a wrapper around configuration, but I think the separation is worth it at the very least for clarity and also for the future where we may do more complex things in response to a secret.

brooksmtownsend (approved):
@protochron and I have discussed the issue I brought up and decided that it would be best to transform manifests of the "old" source/target config format to the new automatically using `wash`. I think it would be nice to return a good error instead of a failed deserialization error / failed validation error if possible, in the case where folks are using the Wadm API directly instead of using wash (e.g. the operator, right?) 

Let's see if we can use a wash plugin for the translation

crates/wadm-types/src/lib.rs
```diff
@@ -271,28 +400,42 @@ impl PartialEq<ConfigProperty> for String {
...
     pub package: String,
     /// WIT interfaces for the link
     pub interfaces: Vec<String>,
     /// Configuration to apply to the source of the link
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub source_config: Vec<ConfigProperty>,
+    pub source: Option<ConfigDefinition>,
     /// Configuration to apply to the target of the link
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub target_config: Vec<ConfigProperty>,
```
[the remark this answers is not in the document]
protochron: Went ahead and updated the code so that both `source_config` and `target_config` transformed into the new representation. We'll have to figure out how to prepare to remove them entirely.

crates/wadm/src/scaler/manager.rs
```diff
@@ -857,6 +934,32 @@ fn config_to_scalers<C: ConfigSource + Send + Sync + Clone>(
...
+    model_name: &str,
+    secrets: &[SecretProperty],
+    policies: &HashMap<&String, &Policy>,
+) -> (Vec<SecretScaler<S>>, Vec<String>) {
+    secrets
+        .iter()
+        .map(|s| {
+            let name = compute_secret_id(model_name, None, &s.name);
+            println!("{:?}", s.source.policy);
+            println!("{:?}", policies);
```
brooksmtownsend, line 948: Let's remove these printlns too

tests/fixtures/manifests/simple.wadm.yaml
```diff
@@ -7,12 +7,27 @@ metadata:
...
       properties:
         image: ghcr.io/wasmcloud/component-http-hello-world:0.1.0
       traits:
         - type: spreadscaler
           properties:
             instances: 1
+
+            # {"name":"", "type": "wasmcloud.dev/secret/v1", key: "", "policy": "\"{\"role-name\":\"test\"}\"", "backend: ""}
+            # {"name":"", "type": "wasmcloud.dev/secret/v1", key: "", "policy": "whatever", "backend: ""}
+            # POLICY_$app-name-$policy-name
```
brooksmtownsend, line 33: Looks like maybe some brainstorming code we had 😄

brooksmtownsend (changes requested):
This PR largely looks really nice. Huge amount of changes, thanks for dealing with the copy/paste. Just a few requests, at least half are to remove comments/printlns

crates/wadm-types/src/lib.rs
```diff
@@ -312,26 +439,120 @@ impl PartialEq<ConfigProperty> for String {
 }
 
 /// Properties for links
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, JsonSchema)]
-#[serde(deny_unknown_fields)]
+#[derive(Debug, Serialize, Clone, PartialEq, Eq, JsonSchema, Default)]
+//#[serde(deny_unknown_fields)]
```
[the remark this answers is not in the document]
protochron: That should be present, fixed

crates/wadm/oam/README.md
```diff
@@ -0,0 +1,78 @@
...
+- `spreadscaler` - Defines the spread of instances of a particular entity across multiple hosts with affinity requirements
+- `link` - A link definition that describes a link between a component and a capability provider or a component and another component
+
+## Example Application YAML
+
+The following is an example YAML file describing an ALC application
+
+```yaml
+apiVersion: core.oam.dev/v1beta1
+kind: Application
```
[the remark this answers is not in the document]
protochron: This was just from the readme at the top level. I removed this readme and the copied json schemas from the crates where I copied them to.

crates/wadm-types/oam/README.md
```diff
@@ -0,0 +1,78 @@
...
+## wasmCloud OAM Traits
+
+The following is a list of the `traits` wasmCloud has added via customization to its application model.
+
+- `spreadscaler` - Defines the spread of instances of a particular entity across multiple hosts with affinity requirements
+- `link` - A link definition that describes a link between a component and a capability provider or a component and another component
+
+## Example Application YAML
+
+The following is an example YAML file describing an ALC application
```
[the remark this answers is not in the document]
protochron: It was just moved from the top level, so I removed it.

crates/wadm-types/oam/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```yaml
+apiVersion: core.oam.dev/v1beta1
+kind: Application
+metadata:
+  name: my-example-app
```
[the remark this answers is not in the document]
protochron: Also removed.

tests/fixtures/manifests/simple.wadm.yaml
protochron: Definitely was. Removed!

brooksmtownsend (approved):
LGTM! Just a test failing it seems

PR #310, opened 2024-06-30
feat(wadm)!: Switch wadm events to limits-based stream and create a new sourcing stream for EventConsumer

joonas:
## Feature or Problem

This switches from our custom mirroring code to [NATS Stream Sourcing](https://docs.nats.io/nats-concepts/jetstream/streams#sources-and-mirrors).

It also changes the wadm events to follow the pattern established for wasmbus events where they are now published under `wadm.evt.<lattice-id>.<event-type>` to make the sourcing consistent.

Finally, it also includes code to clean up any existing `wadm_(multitenant)_mirror` streams on startup so that we don't end up with conflicting subjects (which would prevent the new streams from being created in the first place).

This effectively accomplishes the same thing as #295 using slightly different strategy, while also making the `wadm_events` (and thus turning `wadm.evt.<lattice-id>.<event-type>` into a stream-based API).


src/nats.rs
```diff
@@ -152,6 +152,107 @@ pub async fn ensure_stream(
...
+    description: Option<String>,
+) -> Result<Stream> {
+    let stream_config = StreamConfig {
+        name: name.clone(),
+        description,
+        num_replicas: 1,
+        retention: async_nats::jetstream::stream::RetentionPolicy::Limits,
+        subjects,
+        max_age: DEFAULT_EXPIRY_TIME,
+        storage: async_nats::jetstream::stream::StorageType::File,
```
thomastaylor312, line 168: Nit: Would memory be a better default since it isn't a huge deal if the events drop? It serves more as a collector anyway

src/nats.rs
```diff
@@ -152,6 +152,107 @@ pub async fn ensure_stream(
...
+    if let Ok(stream) = context.get_stream(&name).await {
+        // For now, we only check if the subjects are the same in order to make sure that
+        // newer versions of wadm adjust subjects appropriately. In the case that developers
+        // want to alter the storage or replicas of a stream, for example,
+        // we don't want to override that configuration.
+        if stream.cached_info().config.subjects == stream_config.subjects {
+            return Ok(stream);
+        } else {
+            warn!("Found stream {name} with different configuration, deleting and recreating");
+            context.delete_stream(name).await?;
```
thomastaylor312, line 182: Is this going to cause issues if someone has created their own stream with their requirements (such as more replicas or other things) or if something else is already consuming these downstream? I'd rather just sort the list of subjects and compare that way if possible

src/nats.rs
```diff
@@ -152,6 +152,107 @@ pub async fn ensure_stream(
...
+            let wildcard = match source_stream_subject.starts_with('*') {
+                true => "{{wildcard(2)}}",
+                false => "{{wildcard(1)}}",
+            };
+            let destination = subject.replacen('*', wildcard, 1);
+            subject_transforms.push(SubjectTransform {
+                source: source_stream_subject.clone(),
+                destination,
+            });
+        }
```
thomastaylor312, line 219: Nit: Consider reducing at least this inner for loop to use an `iter` and `map` instead of a for loop. If you're feeling brave, you could also turn the outer loop into an iter as well

src/nats.rs
```diff
@@ -152,6 +152,107 @@ pub async fn ensure_stream(
...
+
+    let stream_config = StreamConfig {
+        name: name.clone(),
+        description,
+        num_replicas: 1,
+        retention: async_nats::jetstream::stream::RetentionPolicy::WorkQueue,
+        subjects: vec![],
+        max_age: DEFAULT_EXPIRY_TIME,
+        sources: Some(sources),
+        storage: async_nats::jetstream::stream::StorageType::File,
```
thomastaylor312, line 236: This definitely only needs to be memory as wadm can recover from heartbeats

thomastaylor312 (changes requested):

src/nats.rs
joonas: I haven't spend any time looking at the implications between the two storage options, but I'm fine with switching this to Memory until we have evidence pointing otherwise 👍

src/nats.rs
joonas: it's funny you say that, because I actually had it implemented that way originally, but it got a little unwieldy to read, so I decided to go with a much simpler (admittedly more verbose) approach.

Happy to bring back the iter approach if we generally prefer that

thomastaylor312: It is just a nit, so no requirement here

src/nats.rs
thomastaylor312: After talking with @joonas offline, turns out there are tradeoffs and configurations needed to use memory storage types. So we're not going to worry about it for now

src/nats.rs
thomastaylor312: https://github.com/wasmCloud/wadm/pull/310#discussion_r1661595119

src/nats.rs
joonas: Addressed this in the latest push.

src/nats.rs
joonas: > Is this going to cause issues if someone has created their own stream with their requirements (such as more replicas or other things) or if something else is already consuming these downstream? 

For downstream this will not create issues, I tested this locally and the net-net on this is that once the Stream is re-created, NATS will continue where it left off, meaning that the only messages that could be lost are the ones that are in the stream exactly at the moment when it's being re-created.

As for letting someone (pre-)create this stream but with different configuration, I'm not sure how easy that would be to support. It feels like if that's the path we want to go down, a better strategy would be to explicitly expose the knobs for configuring the various parts of these stream as flags/environment variables that could be passed down to code that's (re)creating the streams.

WDYT?

thomastaylor312 (approved):

PR #315, opened 2024-07-03
feat(state): retry state updates

thomastaylor312:
This adds retries for putting state into a bucket. It is fairly basic but it gets the job done. We had to add a few more constraints to the `Store` trait to handle the retries (such as cloning data).

### DIFF
--- a/crates/wadm/src/storage/nats_kv.rs
+++ b/crates/wadm/src/storage/nats_kv.rs
@@ -14,12 +14,14 @@
 //! structure in the future
 use std::collections::HashMap;
 use std::io::Error as IoError;
+use std::time::Duration;
 
 use async_nats::{
     jetstream::kv::{Operation, Store as KvStore},
     Error as NatsError,
 };
 use async_trait::async_trait;
+use futures::Future;
 use serde::{de::DeserializeOwned, Serialize};
 use tracing::{debug, error, field::Empty, instrument, trace};
 use tracing_futures::Instrument;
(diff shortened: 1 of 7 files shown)


crates/wadm/src/storage/nats_kv.rs
```diff
@@ -89,6 +91,71 @@ impl NatsKvStore {
...
+                    .internal_list::<T>(lattice_id)
+                    .in_current_span()
+                    .await?;
+                debug!(revision, "Updating data in store");
+                let updated_data = updater(current_data).await?;
+                trace!("Writing bytes to store");
+                // If the function doesn't return any data (such as for deletes), just return early
+                if updated_data.is_empty() {
+                    return Ok(())
+                }
```
brooksmtownsend, line 123: What if we wanted to update the data at key to be an empty vec?

crates/wadm/src/storage/nats_kv.rs
```diff
@@ -89,6 +91,71 @@ impl NatsKvStore {
...
+                match self.store.update(key, updated_data.into(), revision).await {
+                    Ok(_) => return Ok(()),
+                    Err(e) => {
+                        if e.to_string().contains("wrong last sequence") {
+                            debug!(%key, %lattice_id, "Got wrong last sequence when trying to update state. Retrying update operation");
+                            continue;
+                        }
+                        return Err(NatsStoreError::Nats(e.into()));
+                    }
+                    // TODO(thomastaylor312): Uncomment this code once we can update to the latest
```
brooksmtownsend, line 133:
```suggestion
                    // TODO(#316): Uncomment this code once we can update to the latest
```

brooksmtownsend (commented):
One question one suggestion since I filed an issue.

brooksmtownsend (approved):

PR #319, opened 2024-07-12
chore: Add dependabot configuration

joonas:
## Feature or Problem

Helps keep dependencies up-to-date.

I defaulted to `daily` for the interval, but I could see this being something that also gets scheduled for `weekly` cadence, [maybe first thing on Monday morning](https://github.com/wasmCloud/wasmCloud/blob/main/.github/dependabot.yml#L15-L22)?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
I'm good with this, I think I would prefer weekly just to have a batch on Monday but we can always change either way 😄

joonas:
> I'm good with this, I think I would prefer weekly just to have a batch on Monday but we can always change either way 😄

Weekly it is!

Issue #288, opened 2024-05-22
[BUG] Prevent `source_config` for components


brooksmtownsend:
wasmCloud 1.0 does not support `source_config` for components, so we should fail manifest validation if a `source_config` is specified on a component.

PR #337, opened 2024-07-17
feat(validation): validation fails if component has source_config

markkovari:
## Related Issues

#288 

### Unit Test(s)

unit tests should cover the negative case, other validation calls don't have this error

### DIFF
--- a/crates/wadm-types/src/validation.rs
+++ b/crates/wadm-types/src/validation.rs
@@ -298,9 +298,12 @@ pub async fn validate_manifest_file(
 pub async fn validate_manifest_bytes(
     content: impl AsRef<[u8]>,
 ) -> Result<(Manifest, Vec<ValidationFailure>)> {
+    let raw_yaml_content = content.as_ref();
     let manifest =
         serde_yaml::from_slice(content.as_ref()).context("failed to parse manifest content")?;
-    let failures = validate_manifest(&manifest).await?;
+    let mut failures = validate_manifest(&manifest).await?;
+    let mut yaml_issues = validate_raw_yaml(raw_yaml_content)?;
+    failures.append(&mut yaml_issues);
     Ok((manifest, failures))
 }
 
@@ -346,6 +349,14 @@ pub async fn validate_manifest(manifest: &Manifest) -> Result<Vec<ValidationFail
     failures.extend(validate_component_properties(manifest));
     failures.extend(check_duplicate_links(manifest));
     failures.extend(validate_link_configs(manifest));
+    Ok(failures)
+}
+
+pub fn validate_raw_yaml(content: &[u8]) -> Result<Vec<ValidationFailure>> {
+    let mut failures = Vec::new();
+    let raw_content: serde_yaml::Value =
+        serde_yaml::from_slice(content).context("failed read raw yaml content")?;
+    failures.extend(validate_components_configs(&raw_content));
     Ok(failures)
 }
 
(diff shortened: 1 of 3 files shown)


markkovari:
I do have 2 questions tho:
- is it even a valid configuration, if not feel free to point out edge cases to add
- the validation looks similar to a callback hell, hopefully you have wide enough screens, please call me out if it needs refactor

Appreciate

brooksmtownsend (commented):
Hey @markkovari ! I think the structure of this PR looks great, you may just need to consider a change or two after #307. If we want to try and move this forward now, we should at least just change the parsing of the source_config key to be in a link trait

crates/wadm-types/src/validation.rs
```diff
@@ -403,6 +405,33 @@ fn check_dangling_links(manifest: &Manifest) -> Vec<ValidationFailure> {
...
+    let component_traits = components
+        .into_iter()
+        .filter_map(|(name, component)| match &component.traits {
+            Some(traits) => Some((name, traits)),
+            None => None,
+        });
+    for (name, traits) in component_traits {
+        for trait_ in traits {
+            if let TraitProperty::Custom(custom) = trait_.properties.clone() {
+                if let Some(_) = custom.get(forbidden_config_key) {
```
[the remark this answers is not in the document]
markkovari: Cool, thanks for the heads up, we can definitely wait a bit with this than, I was just getting familiar with the project. 🙇

brooksmtownsend:
Alright @markkovari you should be good to proceed with this one 😄 Just take note of the newer structure where the `source` block is the recommended route, and `source_config` is a backwards compatible field. I would recommend just validating that the `source: config:` block isn't present, since we want to deprecate `source_config` out in a future release anyways

markkovari:
Just a question here: since `v0.13.0` the fields are deprecated (`source_config`) and (`target_config`), this is still relevant right?
Currently the yaml parsing into the structure is always [None](https://github.com/wasmCloud/wadm/blob/2c00cada86e0a220cb52405e7a58176b82a93ec7/crates/wadm-types/src/bindings.rs#L501-L502) so it is doable by the casual yaml traversing/parsing method.
And this should be even related to this too:
https://github.com/wasmCloud/wasmCloud/issues/2862

brooksmtownsend:
Hey @markkovari , yep I think it's still relevant here just failing when `source.config` is present instead of `source_config` if that makes sense 😄 

It would be best to do the validation in a place where we can just use the same logic for both

crates/wadm-types/src/validation.rs
```diff
@@ -678,6 +679,43 @@ pub fn validate_component_properties(application: &Manifest) -> Vec<ValidationFa
...
 
+/// Funtion to validate the component configs
+/// from 0.13.0 source_config is deprecated and replaced with source:config:
+/// this function validates the component configs
+pub fn validate_component_configs(application: &Manifest) -> Vec<ValidationFailure> {
+    let mut failures = Vec::new();
+    for component in application.spec.components.iter() {
+        if has_deprecated_trait_property(component) {
+            failures.push(ValidationFailure::new(
+                ValidationFailureLevel::Error,
```
thomastaylor312, line 690: This should be warning level as the field is just deprecated. This would be a change in the test as well to check for warnings

tests/validation.rs
```diff
@@ -118,3 +118,20 @@ async fn validate_policy() -> Result<()> {
...
 }
+
+#[tokio::test]
+async fn validate_deprecated_configs() -> Result<()> {
+    let (_manifest, failures) = validate_manifest_file(
+        "./tests/fixtures/manifests/deprecated-source-and-target-config.yaml",
+    )
+    .await
+    .context("failed to validate manifest")?;
+    assert!(!failures.is_empty(), "failures present");
```
thomastaylor312, line 129: I think this might be left over? You're asserting failure length below

thomastaylor312 (approved):
Almost there @markkovari! I think there is a line in the test left over and the failure should be a warning as it is only a deprecated field

thomastaylor312 (changes requested):
Whoops hit wrong button

thomastaylor312:
This is a great start @markkovari. There is definitely more we could add to this part of validation, but even having something that warns about deprecation is great

tests/validation.rs
```diff
@@ -118,3 +118,20 @@ async fn validate_policy() -> Result<()> {
...
 }
+
+#[tokio::test]
+async fn validate_deprecated_configs() -> Result<()> {
+    let (_manifest, failures) = validate_manifest_file(
+        "./tests/fixtures/manifests/deprecated-source-and-target-config.yaml",
+    )
+    .await
+    .context("failed to validate manifest")?;
+    assert!(!failures.valid(), "expected invalid manifest");
```
thomastaylor312, line 129: Heads up that I think if all of the failures are warnings, it will be a valid manifest

markkovari: Thanks, I just realized that the manifests link traits target_config and source_config keys are not being stored as values. I might be able to do it when storing the original manifest, which is another issue on the repo. Will look into it early next week. 🙇

markkovari:
@thomastaylor312 @brooksmtownsend sorry for the delay
unfortunately the solution is a bit hacky 😅 , but the serialization is always - making the yaml parsed objects link traits - `sounce_config` and `target_config` `None`, not sure if it is possible to do it the "nice way" just like the other steps

Thanks

brooksmtownsend (approved):
LGTM, any thoughts here @thomastaylor312 ? Looks like we can remove the `validate_raw_yaml` bit once we finally remove the source_config/target_config fields

thomastaylor312 (approved):

PR #342, opened 2024-07-22
chore(wadm): use application verbage instead of model

brooksmtownsend:
## Feature or Problem
This PR addresses the verbage in wadm's server handlers module, ensuring that we refer to operations on the wadm API as operations that create/read/update/delete Applications, not Manifests or Models. This gets closer to the verbage we use in wash-cli, and is more intuitive.

I considered splitting the verbs between Application and Manifest, and removing Model, so if any of the usage of Application feels weird let me know.

### DIFF
--- a/crates/wadm/src/server/handlers.rs
+++ b/crates/wadm/src/server/handlers.rs
@@ -164,7 +164,7 @@ impl<P: Publisher> Handler<P> {
                 Err(e) => {
                     self.send_error(
                         msg.reply,
-                        format!("Unable to parse get model request: {e:?}"),
+                        format!("Unable to parse get application request: {e:?}"),
                     )
                     .await;
                     return;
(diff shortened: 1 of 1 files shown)


crates/wadm/src/server/handlers.rs
```diff
@@ -413,7 +413,7 @@ impl<P: Publisher> Handler<P> {
                 Ok(_) => {
                     DeleteModelResponse {
                         result: DeleteResult::Deleted,
-                        message: format!("Successfully deleted model {}", name),
+                        message: format!("Successfully deleted application {}", name),
```
joonas, line 416:
```suggestion
                        message: format!("Successfully deleted application {name}"),
```

joonas (approved):
I'm a fan of this change, I think separating what wadm might call things internally from what they are called in [wasmCloud documentation](https://wasmcloud.com/docs/ecosystem/wadm/) and [`wash`](https://wasmcloud.com/docs/cli/app) makes sense and provides additional clarity/consistency 👍

thomastaylor312 (approved):

protochron (approved):

PR #343, opened 2024-07-23
fix: copy backend key to top level in a generated config value

protochron:
## Feature or Problem
It turns out that it is easier to parse a serialized secrets value if we have the `backend` key at the top level instead of nested in string-encoded JSON. This moves that value out of the policy properties value that the secrets scaler serializes into a config value and instead stores it as a top level key.

## Release Information
0.13.0

### DIFF
--- a/crates/wadm/src/scaler/secretscaler.rs
+++ b/crates/wadm/src/scaler/secretscaler.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{Context, Result};
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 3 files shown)


crates/wadm/src/scaler/secretscaler.rs
```diff
@@ -145,10 +145,16 @@ fn merge_policy_properties(
...
 
-    let properties = PolicyProperties {
+    let mut properties = PolicyProperties {
         policy_type: policy.policy_type.clone(),
         properties: policy.properties.clone(),
     };
+    let backend = properties
+        .properties
+        .remove("backend")
+        .expect("policy does not have a backend property");
```
brooksmtownsend, line 155:
```suggestion
    let backend = properties
        .properties
        .remove("backend")
        .context("policy did not have a backend property")?;
```

brooksmtownsend (changes requested):
Mostly looks good, just a request to bubble up an error instead of an `expect`

protochron:
@brooksmtownsend it did, yeah. I don't know why it isn't being marked as outdated

brooksmtownsend (approved):

Issue #347, opened 2024-07-24
[FEAT] Update the wadm WIT types bindings to support secrets


brooksmtownsend:
Since secrets support landed in the manifests, we've made a few changes in addition to just adding secrets. We've updated the `source_config` and `target_config` blocks to be `source` and `target` instead, and added the `policies` top level block alongside `components`. These need to be reflected in the WIT interfaces.

PR #348, opened 2024-07-24
chore(*): prepare 0.13

brooksmtownsend:
## Feature or Problem
This PR takes care of necessary crate bumps to release wadm 0.13.0, and clients/types 0.2.0. Leaving as draft as I continue to validate the functionality here and take a look at some of the code.

## Related Issues
Note that I've filed #347 since that hasn't been taken care of yet.

### DIFF
--- a/crates/wadm-types/src/lib.rs
+++ b/crates/wadm-types/src/lib.rs
@@ -33,10 +33,11 @@ pub const LINK_TRAIT: &str = "link";
 /// for a manifest
 pub const LATEST_VERSION: &str = "latest";
 
+// TODO: These are copied from the wasmcloud secret-types crate and should just be moved here from there.
 /// The type and version of the secret reference stored as configuration. This is meant as an
 /// internal marker of the format of the serialized secret and should not be referenced by
 /// anything else but wadm. It is intended to help with schema upgrades in the future.
-pub const SECRET_TYPE: &str = "secret-reference.wasmcloud.dev/v1alpha1";
+pub const SECRET_TYPE: &str = "v1.secret.wasmcloud.dev";
 
 /// An OAM manifest
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, utoipa::ToSchema, JsonSchema)]
--- a/crates/wadm-types/src/lib.rs
+++ b/crates/wadm-types/src/lib.rs
@@ -1,6 +1,7 @@
+use std::collections::{BTreeMap, HashMap};
+
 use schemars::JsonSchema;
 use serde::{de, Deserialize, Serialize};
-use std::collections::{BTreeMap, HashMap};
 
 pub mod api;
 #[cfg(feature = "wit")]
@@ -32,12 +33,6 @@ pub const LINK_TRAIT: &str = "link";
 /// The string used for indicating a latest version. It is explicitly forbidden to use as a version
 /// for a manifest
 pub const LATEST_VERSION: &str = "latest";
-
-// TODO: These are copied from the wasmcloud secret-types crate and should just be moved here from there.
-/// The type and version of the secret reference stored as configuration. This is meant as an
-/// internal marker of the format of the serialized secret and should not be referenced by
-/// anything else but wadm. It is intended to help with schema upgrades in the future.
-pub const SECRET_TYPE: &str = "v1.secret.wasmcloud.dev";
 
 /// An OAM manifest
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, utoipa::ToSchema, JsonSchema)]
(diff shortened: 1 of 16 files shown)


crates/wadm-types/src/lib.rs
```diff
@@ -247,9 +247,9 @@ pub struct SecretProperty {
     /// The name of the secret. This is used by a reference by the component or capability to
     /// get the secret value as a resource.
     pub name: String,
-    /// The source of the secret. This indicates how to retrieve the secret value from a secrets
+    /// The properties of the secret that indicate how to retrieve the secret value from a secrets
     /// backend and which backend to actually query.
-    pub source: SecretSourceProperty,
```
brooksmtownsend, line 252: @protochron let me know what you think here, looking at a manifest the `secrets` block was the only place where we had a different name for what is essentially a set of properties, so I figured renaming it here would be beneficial.

tests/fixtures/manifests/simple.wadm.yaml
```diff
@@ -9,20 +9,21 @@ metadata:
 spec:
   policies:
     - name: vault-for-component
-      type: secrets.wasmcloud.dev/v1/vault
+      type: policy.secret.wasmcloud.dev/v1alpha1
       properties:
-        role-name: "test"
-        backend: "vault-prod"
+        role-name: 'test'
+        backend: 'vault-prod'
```
joonas, line 15: If we want to stay true to what's implemented, this would be a good representation:
```suggestion
        backend: 'vault-prod'
        role_name: 'test'
        mount_path: 'jwt'
```

crates/wadm-types/src/validation.rs
```diff
@@ -520,31 +522,41 @@ fn check_dangling_links(manifest: &Manifest) -> Vec<ValidationFailure> {
...
+                    if policy.policy_type != SECRET_POLICY_TYPE {
+                        failures.push(ValidationFailure::new(
+                            ValidationFailureLevel::Error,
+                            format!(
+                                "policy '{}' type did not match required type '{}'",
+                                policy.name, SECRET_POLICY_TYPE
+                            ),
+                        ));
+                    }
+                }
```
protochron, line 552: A few things:
* we will likely have other policies that wadm doesn't know about, so we shouldn't have a validation failure if the policy type isn't a secret policy
* The `backend` key should only be checked if policy.policy_type != SECRET_POLICY_TYPE`

protochron (changes requested):

crates/wadm-types/src/validation.rs
brooksmtownsend: Great catch. I thought this was only called from the secret context, let me fix this.

brooksmtownsend:
Note: I brought in the wasmcloud-secrets-types crate so this will fail for the moment until https://github.com/wasmCloud/wasmCloud/pull/2564 merges & I can cut a release of the secrets types crate.

crates/wadm-types/src/lib.rs
```diff
@@ -33,10 +33,11 @@ pub const LINK_TRAIT: &str = "link";
 /// for a manifest
 pub const LATEST_VERSION: &str = "latest";
 
+// TODO: These are copied from the wasmcloud secret-types crate and should just be moved here from there.
 /// The type and version of the secret reference stored as configuration. This is meant as an
 /// internal marker of the format of the serialized secret and should not be referenced by
 /// anything else but wadm. It is intended to help with schema upgrades in the future.
-pub const SECRET_TYPE: &str = "secret-reference.wasmcloud.dev/v1alpha1";
+pub const SECRET_TYPE: &str = "v1.secret.wasmcloud.dev";
```
protochron, line 40: This should probably be `secret.wasmcloud.dev/v1` to match how we're handling things elsewhere

protochron (changes requested):

crates/wadm-types/src/lib.rs
joonas: should it not be `v1alpha1`?

brooksmtownsend: I don't know how I left this but will address

thomastaylor312 (approved):

crates/wadm/src/scaler/secretscaler.rs
```diff
@@ -19,43 +15,54 @@ use crate::{
     workers::SecretSource,
 };
 
-pub const SECRET_CONFIG_PREFIX: &str = "SECRET";
+use super::compute_id_sha256;
 
 pub struct SecretScaler<SecretSource> {
     secret_source: SecretSource,
-    id: String,
+    /// The key to use in the configdata bucket for this secret
```
joonas, line 22: Is this supposed to be above `id`?

protochron (approved):

PR #352, opened 2024-07-26
fix: remove serial from tests

ricochet:
Remove serial now that the doc'd bug for why serial has been fixed.

Additionally use project scoped paths instead of relative paths to enable easier gdb and vscode debugging.


brooksmtownsend (approved):

joonas:
I think all of the changes being proposed here got taken care of as part of #365 and #369

brooksmtownsend:
Closing as-per above where this is already taken care of!

Issue #345, opened 2024-07-24
[BUG] Stopping the last host in a lattice results in error messages


brooksmtownsend:
If you stop the last host in the lattice, wadm will receive the host_stopped event and attempt to do a reconciliation pass. This results in the following messages:

```
2024-07-24T14:17:29.728699Z ERROR observe{subscribe_topics=["wasmbus.evt.*.>"]}:consumer_worker{topic=wadm_event_consumer.evt.default.> worker_type=wadm::workers::event::EventWorker<wadm::storage::nats_kv::NatsKvStore, wasmcloud_control_interface::client::Client, async_nats::jetstream::context::Context>}: wadm::consumers::manager: Got error from worker error=Other("Did not receive a response to get links: no responders: no responders")
2024-07-24T14:17:29.730616Z ERROR observe{subscribe_topics=["wasmbus.evt.*.>"]}:consumer_worker{topic=wadm_event_consumer.evt.default.> worker_type=wadm::workers::event::EventWorker<wadm::storage::nats_kv::NatsKvStore, wasmcloud_control_interface::client::Client, async_nats::jetstream::context::Context>}: wadm::consumers::manager: Got error from worker error=Other("Did not receive a response to get links: no responders: no responders")
2024-07-24T14:17:29.732251Z ERROR observe{subscribe_topics=["wasmbus.evt.*.>"]}:consumer_worker{topic=wadm_event_consumer.evt.default.> worker_type=wadm::workers::event::EventWorker<wadm::storage::nats_kv::NatsKvStore, wasmcloud_control_interface::client::Client, async_nats::jetstream::context::Context>}: wadm::consumers::manager: Got error from worker error=Other("Did not receive a response to get links: no responders: no responders")
```

This is because wadm attempts to query the links, but no host is available to answer that query. If a host stops we should attempt to reconcile to find where components and providers need to rebalance, but not query links (or at least not fail when querying them) as this is not an error case we need to be concerned about.

PR #360, opened 2024-08-01
fix(#345): use cached links when req fails

brooksmtownsend:
## Feature or Problem
This PR changes the snapshot store to use cached links in the case of the last host in the lattice stopping, preventing 3 error logs & a failure to reconcile.

## Related Issues
#345

## Release Information
next patch release of wadm

### DIFF
--- a/crates/wadm/src/storage/snapshot.rs
+++ b/crates/wadm/src/storage/snapshot.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use tokio::sync::RwLock;
+use tracing::debug;
 use wasmcloud_control_interface::InterfaceLinkDefinition;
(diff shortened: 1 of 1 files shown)


crates/wadm/src/storage/snapshot.rs
```diff
@@ -86,7 +87,14 @@ where
             .into_iter()
             .map(|(key, val)| (key, serde_json::to_value(val).unwrap()))
             .collect::<HashMap<_, _>>();
-        let links = self.lattice_source.get_links().await?;
+
+        // If we fail to get the links, that likely just means the lattice source is down, so we
+        // just what we have cached
```
joonas, line 92:
```suggestion
        // If we fail to get the links, that likely just means the lattice source is down, so we
        // just fall back on what we have cached
```

joonas (approved):

PR #367, opened 2024-08-02
chore: Replace wash up with testcontainers

### LINKED ISSUE #353: [BUG] flake due to dependency on download
This is an integration test that's flaky in [CI](https://github.com/wasmCloud/wadm/actions/runs/10115136397/job/27975239328?pr=352) and for first-time users.

Steps to reproduce:
```bash
# clear downloads
rm -rf ~/.wash/downloads
cargo test --package wadm-cli --test command_worker_integration -- test_annotation_stop --exact --show-output
```

The error, if the test flakes, will be:
```bash
 ---- test_annotation_stop stdout ----
thread 'test_annotation_stop' panicked at tests/helpers.rs:120:37:
Unable to run detached wash up: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error stopping wasmcloud host: No such file or directory (os error 2)
```

This is from the host download taking time to download.

Where possible we should avoid downloads including in integration tests. But if we need to stage a pipeline to run more e2e style tests, then the integration tests should be staged prior to triggering them like with a setup action in CI.

joonas:
## Feature or Problem

This replaces a bunch of the `wash up` based test setup with [`testcontainers`](https://github.com/testcontainers/testcontainers-rs/) based setup instead.

## Related Issues

Closes #353

### DIFF
--- a/tests/command_worker_integration.rs
+++ b/tests/command_worker_integration.rs
@@ -9,8 +9,8 @@ use wadm::{
 
 mod helpers;
 use helpers::{
-    setup_test_wash, StreamWrapper, TestWashConfig, HELLO_COMPONENT_ID, HELLO_IMAGE_REF,
-    HTTP_SERVER_COMPONENT_ID, HTTP_SERVER_IMAGE_REF,
+    setup_env, StreamWrapper, HELLO_COMPONENT_ID, HELLO_IMAGE_REF, HTTP_SERVER_COMPONENT_ID,
+    HTTP_SERVER_IMAGE_REF,
 };
 
 #[tokio::test]
@@ -18,13 +18,12 @@ use helpers::{
 // note this test should probably be changed to an e2e test as the order of events is somewhat flaky
 #[serial]
 async fn test_commands() {
-    let config = TestWashConfig::random().await.unwrap();
-    let _guard = setup_test_wash(&config).await;
+    let env = setup_env().await;
+    let nats_client = env.nats_client().await;
 
-    let mut wrapper = StreamWrapper::new("commands_integration".into(), config.nats_port).await;
+    let mut wrapper = StreamWrapper::new("commands_integration".into(), nats_client.clone()).await;
 
-    let ctl_client =
-        wasmcloud_control_interface::ClientBuilder::new(wrapper.client.clone()).build();
+    let ctl_client = wasmcloud_control_interface::ClientBuilder::new(nats_client.clone()).build();
     let worker = CommandWorker::new(ctl_client.clone());
 
     let host_id = ctl_client
@@ -40,8 +39,7 @@ async fn test_commands() {
         .id
         .to_owned();
 
-    let mut sub = wrapper
-        .client
+    let mut sub = nats_client
         .subscribe("wasmbus.evt.default.>".to_string())
         .await
         .unwrap();
@@ -328,17 +326,15 @@ async fn test_commands() {
 #[serial]
 async fn test_annotation_stop() {
     // This test is a sanity check that we only stop annotated components
-    let config = TestWashConfig::random().await.unwrap();
-    let _guard = setup_test_wash(&config).await;
+    let env = setup_env().await;
+    let nats_client = env.nats_client().await;
 
-    let mut wrapper = StreamWrapper::new("annotation_stop".into(), config.nats_port).await;
+    let mut wrapper = StreamWrapper::new("annotation_stop".into(), nats_client.clone()).await;
 
-    let ctl_client =
-        wasmcloud_control_interface::ClientBuilder::new(wrapper.client.clone()).build();
+    let ctl_client = wasmcloud_control_interface::ClientBuilder::new(nats_client.clone()).build();
     let worker = CommandWorker::new(ctl_client.clone());
 
-    let mut sub = wrapper
-        .client
+    let mut sub = nats_client
         .subscribe("wasmbus.evt.default.>".to_string())
         .await
         .unwrap();
@@ -479,7 +475,7 @@ async fn test_annotation_stop() {
 
 async fn wait_for_event(sub: &mut async_nats::Subscriber, match_text: &str) {
     // Providers can take a bit to start if they are downloading
-    let mut interval = tokio::time::interval(std::time::Duration::from_secs(15));
+    let mut interval = tokio::time::interval(std::time::Duration::from_secs(30));
     // Consume the initial tick
     interval.tick().await;
     loop {
--- a/tests/command_worker_integration.rs
+++ b/tests/command_worker_integration.rs
@@ -18,8 +18,13 @@ use helpers::{
 // note this test should probably be changed to an e2e test as the order of events is somewhat flaky
 #[serial]
 async fn test_commands() {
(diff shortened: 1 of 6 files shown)


tests/command_worker_integration.rs
```diff
@@ -479,7 +475,7 @@ async fn test_annotation_stop() {
 
 async fn wait_for_event(sub: &mut async_nats::Subscriber, match_text: &str) {
     // Providers can take a bit to start if they are downloading
-    let mut interval = tokio::time::interval(std::time::Duration::from_secs(15));
+    let mut interval = tokio::time::interval(std::time::Duration::from_secs(30));
```
joonas, line 478: I bumped this up to 30, because I noticed it was fail reasonably often with 15, but not once at 30 in my testing.

It's of course not very scientific, but I think having to wait a bit longer is a reasonable trade off to having to re-run the entire suite.

tests/helpers.rs
```diff
@@ -181,19 +125,22 @@ pub async fn wait_for_server(url: &str) {
...
         panic!(
-            "wash command didn't exit successfully: {}",
-            String::from_utf8_lossy(&output.stderr)
-        )
+            "wash command ({:?}) didn't exit successfully: {}",
+            args.into_iter()
+                .map(|i| i.as_ref().to_str().unwrap_or_default().to_owned())
+                .collect::<Vec<String>>(),
+            String::from_utf8_lossy(&output.stderr),
+        );
```
joonas, line 143: I'm also wondering if this should actually return a `Result` instead so that you could localize the error where it happens, rather than awkwardly getting a reference to the `panic!(...)` when it fails.

joonas:
I believe the changes from https://github.com/wasmCloud/wasmCloud/pull/2667 should fix the error in tests

tests/event_consumer_integration.rs
```diff
@@ -303,7 +303,8 @@ async fn test_nack_and_rereceive() -> Result<()> {
         "--ctl-port",
         &ctl_port,
     ])
-    .await;
+    .await
+    .unwrap();
```
brooksmtownsend, line 307: nit: this could be an `expect` like the other wash helpers

tests/helpers.rs
```diff
@@ -31,132 +33,74 @@ fn get_random_tcp_port() -> u16 {
...
-                Err(e) => eprintln!("Error stopping wasmcloud host: {e}"),
-                _ => (),
-            }
-        }
+impl TestEnv {
+    pub async fn nats_port(&self) -> u16 {
+        self.nats_server
+            .get_host_port_ipv4(DEFAULT_NATS_PORT)
+            .await
+            .unwrap()
```
brooksmtownsend, line 46: This feels like a good place for an expect

tests/helpers.rs
```diff
@@ -31,132 +33,74 @@ fn get_random_tcp_port() -> u16 {
...
-    /// Build a test wash configuration with randomized ports
-    pub async fn random() -> Result<TestWashConfig> {
-        let nats_port = Some(get_random_tcp_port());
-
-        Ok(TestWashConfig {
-            nats_port,
-            ..TestWashConfig::default()
-        })
+    pub async fn nats_url(&self) -> String {
+        let nats_host = self.nats_server.get_host().await.unwrap();
```
brooksmtownsend, line 50: Same thing re: expect

tests/helpers.rs
```diff
@@ -31,132 +33,74 @@ fn get_random_tcp_port() -> u16 {
...
+        let nats_port = self.nats_port().await;
+        format!("{nats_host}:{nats_port}")
     }
 
-    /// Get the NATS URL for this config
-    pub fn nats_url(&self) -> String {
-        format!("127.0.0.1:{}", self.nats_port.unwrap_or(DEFAULT_NATS_PORT))
+    pub async fn nats_client(&self) -> async_nats::Client {
+        async_nats::connect(self.nats_url().await).await.unwrap()
```
brooksmtownsend, line 56: I think this would do well as a result that we return, since the connection is more likely to fail than other things

tests/helpers.rs
```diff
@@ -31,132 +33,74 @@ fn get_random_tcp_port() -> u16 {
...
-        "--host-seed",
-        &seed,
-        "--wasmcloud-version",
-        "v1.0.4",
-    ]);
-    if cfg.nats_connect_only {
-        args.push("--nats-connect-only");
+pub async fn setup_env() -> TestEnv {
+    let nats_server = start_nats_server().await.unwrap();
+    let bridge_nats_server_ip = nats_server.get_bridge_ip_address().await.unwrap();
```
brooksmtownsend, line 62: expect

tests/helpers.rs
```diff
@@ -31,132 +33,74 @@ fn get_random_tcp_port() -> u16 {
...
+    let wasmcloud_host = start_wasmcloud_host(TestWasmCloudHostConfig {
+        nats_ip: bridge_nats_server_ip.to_string(),
+        nats_port: DEFAULT_NATS_PORT.to_string(),
+        wasmcloud_version: "1.1.0".to_string(),
+    })
+    .await
+    .unwrap();
```
brooksmtownsend, line 69: This would be a good expect to give context

brooksmtownsend (changes requested):
This looks great! The testcontainers looks real simple to use.

I have a bunch of requests for unwraps to be `expect`s so we can make sure we're bubbling up context. Additionally, I think it would be great to either update the `Makefile` to no longer run the docker container manually, or just remove it in favor of using cargo test.

tests/event_consumer_integration.rs
joonas: Good call, thanks!

tests/helpers.rs
joonas: I ended up returning the error up to the caller so that it can be surfaced appropriately in the calling code

tests/helpers.rs
joonas: Same re: returning the error to the caller and handling it closer to the code that's calling in.

tests/helpers.rs
joonas: Same re: returning the error to the caller and handling it closer to the code that's calling in.

tests/helpers.rs
joonas: Same re: returning the error to the caller and handling it closer to the code that's calling in.

joonas:
> I have a bunch of requests for unwraps to be `expect`s so we can make sure we're bubbling up context. Additionally, I think it would be great to either update the `Makefile` to no longer run the docker container manually, or just remove it in favor of using cargo test.

I believe there are other tests where the dockerized nats-server is used, so I think that's still valid setup for the time being, and the good news is that this does not in any way interfere with that (as evidenced by the tests passing).

I would love to see the entire test suite moved over to this approach instead, because it provides isolation at per-test level thus making it possible to run multiple tests that might be using the same subjects to test at the same time, but that wasn't in scope for this particular change, that would build on these changes.

brooksmtownsend (approved):

PR #368, opened 2024-08-02
feat(client)!: return name and version from deploy model

brooksmtownsend:
## Feature or Problem
This PR is a simple fix to return the name and version of a deployed manifest with the DeployModelResponse, which assists with the data that clients can use to know what version was deployed even if one wasn't supplied.

### DIFF
--- a/crates/wadm/src/server/handlers.rs
+++ b/crates/wadm/src/server/handlers.rs
@@ -500,6 +500,8 @@ impl<P: Publisher> Handler<P> {
                         serde_json::to_vec(&DeployModelResponse {
                             result: DeployResult::NotFound,
                             message: format!("Application with the name {name} not found"),
+                            name: name.to_string(),
+                            version: req.version.clone(),
                         })
                         .unwrap_or_default(),
                     )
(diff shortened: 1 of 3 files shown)


crates/wadm/src/server/handlers.rs
```diff
@@ -750,19 +764,25 @@ impl<P: Publisher> Handler<P> {
                 .map(|_| DeployModelResponse {
                     result: DeployResult::Acknowledged,
                     message: format!("Successfully undeployed application {name}"),
+                    name: name.to_string(),
+                    version: deployed_version.clone().unwrap_or_default(),
```
joonas, line 768: Isn't this potentially a bit misleading to send back a version when you've undeployed something, because by definition it is not in fact deployed.

Perhaps it's splitting hairs over semantics though and it would. be better to always send back a version regardless.

joonas (approved):

crates/wadm/src/server/handlers.rs
brooksmtownsend: Potentially, yes. I could set the version to an option to account for this 🤔

brooksmtownsend: @joonas changed to an option!

joonas (approved):
Looks good to me! 👍

PR #369, opened 2024-08-02
chore: Migrate more tests over to using testcontainers for setup

joonas:
## Feature or Problem

Since #367 introduced a `testcontainers`-based setup for the test environment, which provides some nice properties over running a shared NATS server, we should adopt it for more tests.


brooksmtownsend (approved):
Looks great!

PR #381, opened 2024-08-09
feat: support shared components and providers

### LINKED ISSUE #277: [FEAT] Support sharing components/providers between manifests
Just like named configuration can be specified solely by name, without specifying the properties, and sourced from outside of an application, it would be great to be able to share a component or capability provider between different applications to cut down on duplication of compute.

For example, a minimal HTTP component manifest:
```yaml
kind: Application
metadata:
  name: rust-hello-world
  annotations:
    version: v0.0.1
    description: "HTTP hello world demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)"
spec:
  components:
    - name: http-component
      type: component
      properties:
        image: file://./build/http_hello_world_s.wasm
      traits:
        # Govern the spread/scheduling of the component
        - type: spreadscaler
          properties:
            replicas: 1

    # Add a capability provider that enables HTTP access
    - name: httpserver
      type: capability
      properties:
        image: ghcr.io/wasmcloud/http-server:0.20.0
      traits:
        # Link the httpserver to the component, and configure the HTTP server
        # to listen on port 8080 for incoming requests
        - type: link
          properties:
            target: http-component
            namespace: wasi
            package: http
            interfaces: [incoming-handler]
            source_config:
              - name: default-http
                properties:
                  address: 127.0.0.1:8080
```

If you imagine we have two of these basic HTTP component YAMLs, most of the configuration will look the same but we'll end up using two HTTP servers to facilitate communication.

To throw out a strawman, one possible solution to this would be to reference a provider by name and, optionally, manifest:
```yaml
kind: Application
metadata:
  name: managed
  annotations:
    version: v0.0.1
    description: "Managed shared provider"
spec:
  components:
    - name: httpserver
      type: capability
      properties:
        image: ghcr.io/wasmcloud/http-server:0.20.0
```

Then, in our updated application manifest:
```yaml
kind: Application
metadata:
  name: rust-hello-world
  annotations:
    version: v0.0.1
    description: "HTTP hello world demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)"
spec:
  components:
    - name: http-component
      type: component
      properties:
        image: file://./build/http_hello_world_s.wasm
      traits:
        # Govern the spread/scheduling of the component
        - type: spreadscaler
          properties:
            replicas: 1

    # Reuse a capability from a different manifest
    - name: httpserver
      type: capability
      properties:
        app: managed
        component: httpserver
      traits:
        # Link the httpserver to the component, and configure the HTTP server
        # to listen on port 8080 for incoming requests
        - type: link
          properties:
            target: http-component
            namespace: wasi
            package: http
            interfaces: [incoming-handler]
            source_config:
              - name: default-http
                properties:
                  address: 127.0.0.1:8080
```

The benefit of only allowing references to other applications is validation, which would let us look up dependent providers before deploying an application (and to warn when undeploying the `managed` application)

brooksmtownsend:
## Feature or Problem
This PR adds support to wadm for two major changes:
1. Allow applications to attach a special `wasmcloud.dev/shared: 'true'` annotation, which marks an application as shared to the entire lattice
2. Allow applications to link components and providers to/from components in a different deployed application that is shared

Components and capability providers can now specify a shared application in their properties rather than supplying an image reference. When an application is deployed that references a component in a different application, wadm will validate that the application is deployed and that the referenced component exists.

You cannot use the shared component feature to alter configuration, identifiers, or scaling properties of a shared component. I could see a feature in the future to add additional scalers, but it might cause undesirable behavior and it would be generally just better to supply your own component instead of using a shared one at this point.

This PR is getting really large, so I'm going to push off additional validation in the handlers to a separate PR that can be reviewed independently.

Here are things that should be done before this feature lands, but based on their checked status aren't completed in this PR:

(Captured these in #451)
- [ ] Ensure applications that have dependents can't undeploy or delete
- [ ] Ensure applications that are shared can't change to be unshared if there are dependents
- [ ] Ensure applications that are shared can't remove the shared components if there are dependents of a component
- [ ] Surface shared component status in applications that depend on that component (this is hard and I'm currently thinking of the best way to do this)
- [x] Double check that applications can't be deployed with both image/application reference or neither
- [x] Update WIT representation of the component/capability properties

## Related Issues
Fixes #277

## Release Information
wadm 0.18

This feature will be released with 0.18.0 and denoted an **experimental** feature for wadm. The cross-manifest linking works well and is e2e tested. There are additional guardrails we may want to introduce here, or change the way the feature works.

## Consumer Impact
This doesn't include any breaking changes for existing manifests

### Unit Test(s)
I'd like to include additional unit tests in the `convert.rs` file that prove that the conversion between a manifest + scalers is properly handled, given how critical and complex that code is. This is made harder by the need for a bunch of generics

### Acceptance or Integration
Added the `e2e_shared` integration test suite that tests sharing components, providers, and multiple aspects of linking to/from shared components.

- [x] I'd also like to include some failure scenarios in this integration test suite.


crates/wadm-types/src/validation.rs
```diff
@@ -596,6 +597,76 @@ fn validate_policies(manifest: &Manifest) -> Vec<ValidationFailure> {
...
+                // could specify additional config and actually overwrite the original manifest's config.
+                (None, Some(shared_properties)) if !config.is_empty() => {
+                    failures.push(ValidationFailure::new(
+                        ValidationFailureLevel::Error,
+                        format!(
+                            "Shared component '{}' cannot specify additional 'config'",
+                            shared_properties.name
+                        ),
+                    ));
+                }
```
joonas, line 647: Related, would specifying `secrets` here be a problem? Or perhaps not a problem, but create unnecessary work for the host when it won't actually be used by the shared component?

joonas:
I'm not sure that it necessarily makes sense to address in this PR, but something that I think needs to be addressed before this feature makes it into a release / becomes available for use is the reporting from a shared application/component's perspective.

If I'm providing a shared component to the lattice, I would like to be able to query what all applications are linking against my shared component so that I can have a conversation with the teams responsible for those applications about the changes I may need to make.

As I currently understand things, this change works well from the perspective of establishing a link to a shared component, but as of right now there is nothing to help the teams providing these shared components understand their "dependents", which is (imo) an absolute necessity for any platform-ish teams to be able to successfully operate.

The other thing that I'm not quite clear on is how would an upgrade to/of a shared component be facilitated, would you be forced to spin up a new application (or perhaps add a new shared component to your existing application) with a different/newer version and then encourage the internal teams to upgrade?

crates/wadm-types/src/validation.rs
brooksmtownsend: Oh yes, at least here it would be just fully ignored so we should report this as an error

brooksmtownsend:
> As I currently understand things, this change works well from the perspective of establishing a link to a shared component, but as of right now there is nothing to help the teams providing these shared components understand their "dependents", which is (imo) an absolute necessity for any platform-ish teams to be able to successfully operate.

Yeah I agree, I captured that in #451 as a followup since this is too hard (imo) to find out just by querying NATS

> The other thing that I'm not quite clear on is how would an upgrade to/of a shared component be facilitated, would you be forced to spin up a new application (or perhaps add a new shared component to your existing application) with a different/newer version and then encourage the internal teams to upgrade?

Yeah so here, right now, there's no validation rules and the expectation is you could updated a shared component to a newer version just fine. It will be up to the deployer to make sure they aren't breaking dependents. I don't think that this is ideal, but I worry that overprescribing deployment rules will make using this feature harder.

crates/wadm-types/src/bindings.rs
```diff
@@ -5,9 +5,10 @@ use crate::{
     },
     CapabilityProperties, Component, ComponentProperties, ConfigDefinition, ConfigProperty,
     LinkProperty, Manifest, Metadata, Policy, Properties, SecretProperty, SecretSourceProperty,
-    Specification, Spread, SpreadScalerProperty, TargetConfig, Trait, TraitProperty,
+    SharedApplicationComponentProperties, Specification, Spread, SpreadScalerProperty,
+    TargetConfig, Trait, TraitProperty,
 };
-use wasmcloud::wadm;
+use wasmcloud::wadm::{self};
```
thomastaylor312, line 11: This is likely going to trigger a clippy lint and should just be `wasmcloud::wadm`

crates/wadm-types/src/lib.rs
```diff
@@ -67,11 +69,65 @@ impl Manifest {
...
+    }
+
     /// Returns the components in the manifest
     pub fn components(&self) -> impl Iterator<Item = &Component> {
         self.spec.components.iter()
     }
 
+    /// Helper function to find shared components that are missing from the given list of
+    /// deployed applications
+    pub fn missing_shared_components(&self, deployed_apps: &Vec<&Manifest>) -> Vec<&Component> {
```
thomastaylor312, line 87: Nit: This should probably just be `&[&Manifest]` instead. Generally clippy warns on this too

crates/wadm/src/scaler/convert.rs
```diff
@@ -0,0 +1,780 @@
...
+    daemonscaler::{provider::ProviderDaemonScaler, ComponentDaemonScaler},
+    secretscaler::SecretScaler,
+    spreadscaler::{
+        link::{LinkScaler, LinkScalerConfig},
+        provider::{ProviderSpreadConfig, ProviderSpreadScaler},
+    },
+    BackoffWrapper,
+};
+
+pub type BoxedScaler = Box<dyn Scaler + Send + Sync + 'static>;
```
thomastaylor312, line 38: Nit: These should both have doc comments since they're `pub`

crates/wadm/src/scaler/convert.rs
```diff
@@ -0,0 +1,780 @@
...
+/// Helper function, primarily to remove nesting, that extends a [`ScalerList`] with all scalers
+/// from a (Wasm) component [`Component`]
+///
+/// # Arguments
+/// * `scalers` - The list of scalers to extend
+/// * `components` - The list of components to convert
+/// * `properties` - The properties of the component to convert
+/// * `traits` - The traits of the component to convert
+/// * `manifest_name` - The name of the manifest that the scalers are being created for
+/// * `application_name` - The name of the application that the component is in
```
thomastaylor312, line 156: Nit: It might be worth noting that this arg should be the name of the "source" application when dealing with something that's shared. I figured it out, but just trying to think of anyone else looking at this down the line

crates/wadm/src/scaler/convert.rs
```diff
@@ -0,0 +1,780 @@
...
+
+/// Helper function, primarily to remove nesting, that extends a [`ScalerList`] with all scalers
+/// from a capability provider [`Component`]
+/// /// # Arguments
+/// * `scalers` - The list of scalers to extend
+/// * `components` - The list of components to convert
+/// * `properties` - The properties of the capability provider to convert
+/// * `traits` - The traits of the component to convert
+/// * `manifest_name` - The name of the manifest that the scalers are being created for
+/// * `application_name` - The name of the application that the component is in
```
thomastaylor312, line 311: Same nit here re: shared stuff

crates/wadm/src/scaler/statusscaler.rs
```diff
@@ -0,0 +1,66 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use wadm_types::{api::StatusInfo, TraitProperty};
+
+use crate::{commands::Command, events::Event, scaler::Scaler};
+
+/// The StatusScaler is a scaler that only reports a predefined status and does not perform any actions.
```
thomastaylor312, line 7: I appreciate this clever trick

tests/e2e_shared.rs
```diff
@@ -0,0 +1,477 @@
...
+        .await;
+    client_info.add_wadm_client(SHARED_PROVIDERS_LATTICE).await;
+    client_info.add_ctl_client(INVALID_TEST_LATTICE, None).await;
+    client_info.add_wadm_client(INVALID_TEST_LATTICE).await;
+    client_info.launch_wadm().await;
+
+    // Wait for the first event on the lattice prefix before we start deploying and checking
+    // statuses. Wadm can absolutely handle hosts starting before you start the wadm process, but the first event
+    // on the lattice will initialize the lattice monitor and for the following test we quickly assert things.
+    let mut sub = client_info
```
thomastaylor312, line 53: Nit: Should we pull out the setup stuff here (which I think you copied) into a common helper for the tests?

thomastaylor312 (approved):
Approved with some nits

crates/wadm-types/src/lib.rs
brooksmtownsend: This gave me some issue w/ collecting into a Sized collection /shrug

crates/wadm/src/scaler/convert.rs
brooksmtownsend: Maybe they shouldn't be 🤔 Will add doc comments if they do need to be

tests/e2e_shared.rs
brooksmtownsend: Agreed! I added an issue for this https://github.com/wasmCloud/wadm/issues/456

Issue #278, opened 2024-05-07
[FEAT] `wadm.api.{lattice}.model.list` should return the response wrapper type


lachieh:
At the moment, requests on the `wadm.api.{lattice}.model.list` topic return a list of model summaries. All the other api topics return a response wrapper with the shape below. The `list` operation should return the same wrapper.

```
{
   status: 'success' | 'error',
   message: string,
   ...
}
```

lachieh:
Rather than changing the existing api, @brooks suggested that we do the k8s thing and add a `model.get` in addition to `model.get.<app_name>` that responds with the expected wrapper type and deprecate the `model.list` topic.

lachieh:
not stale

lachieh:
Closed by #400

PR #400, opened 2024-08-28
fix(server): deprecate and replace model.list operation with model.get

ahmedtadde:
## Feature or Problem
fix response object for `model.list` operation so that it is consistent with response objects of other operations.

## Related Issues
#278 

## Consumer Impact
deprecated and existing model operation. `model.list` still works but (1) has a deprecation warning message & returns an updated response object that is not compatible with current/previous one.

## Testing
all existing tests pass. no regression

### DIFF
--- a/crates/wadm-types/src/api.rs
+++ b/crates/wadm-types/src/api.rs
@@ -23,6 +23,14 @@ pub struct GetModelResponse {
     pub manifest: Option<Manifest>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ListModelsResponse {
+    pub result: GetResult,
+    #[serde(default)]
+    pub message: String,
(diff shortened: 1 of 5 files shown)


brooksmtownsend (approved):
Beautiful 🚀 We should change this in the wash CLI as well

ahmedtadde:
> Beautiful 🚀 We should change this in the wash CLI as well

so what should be the order of operations? I tried opening a PR for [the wash-cli change](https://github.com/wasmCloud/wasmCloud/issues/2884) but then I saw that `wash-cli/wash-lib` is using `wadm v0.11`. so , I am guessing we merge this and then update the `wash-cli` side once it updgrades its `wadm` version dependency.

if we merge this without making any changes in `wash-cli` then

PR #408, opened 2024-09-02
chore: Helm chart reconfiguration

lxfontes:
# Changing `wadm` and `config` positions in values.yaml

When bringing in wadm as a subchart, the resulting values.yaml stutters. Ex:
```
nats:
  config:
   ...

wadm:
  wadm:
    config:
```

Further example in https://github.com/wasmCloud/wasmCloud-contrib/pull/19

Increasing release from `0.2.5` to `0.3.0` due to this config change.

# Templating in `nats.server`

Allows dynamic values, helpful in automated installs. Example:

```
config:
  wadm:
   nats:
    server: "{{ .Release.Name }}"
```

### DIFF
--- a/charts/wadm/values.yaml
+++ b/charts/wadm/values.yaml
@@ -1,4 +1,4 @@
-wadm:
+config:
   # replicas represents the number of copies of wadm to run
   replicas: 1
   # image represents the image and tag for running wadm
@@ -7,7 +7,7 @@ wadm:
     pullPolicy: IfNotPresent
     # Overrides the image tag whose default is the chart appVersion.
     tag: ""
-  config:
+  wadm:
     apiPrefix: ""
     streamPrefix: ""
     cleanupInterval: ""
@@ -48,7 +48,8 @@ serviceAccount:
 podAnnotations: {}
 podLabels: {}
 
-podSecurityContext: {}
+podSecurityContext:
+  {}
   # fsGroup: 1000
 
 securityContext:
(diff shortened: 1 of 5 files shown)


joonas (changes requested):
I'd like to drop the `wadm` key from config entirely, adding that in this chart was a mistake in the first place, it really needs to be left up to the charts to that might include this chart as a subchart.

I think the ⚖️ here is that there are a bunch of examples out there (docs, other repos) that reference this current format, which means that if we go ahead and drop the key, we'll want to make sure that all of those references are also updated with it.

The same would apply if we went ahead and renamed the key this way as well, but it's something to keep in mind.

charts/wadm/values.yaml
```diff
@@ -48,7 +48,8 @@ serviceAccount:
 podAnnotations: {}
 podLabels: {}
 
-podSecurityContext: {}
+podSecurityContext:
+  {}
```
joonas, line 52: Can you revert this change, I'm pretty sure we want to keep it the way it was.

charts/wadm/values.yaml
```diff
@@ -62,8 +63,8 @@ securityContext:
   seccompProfile:
     type: "RuntimeDefault"
 
-
-resources: {}
+resources:
+  {}
```
joonas, line 67: Can you revert this change, I'm pretty sure we want to keep it the way it was.

joonas (changes requested):

charts/wadm/values.yaml
lxfontes: no problem. heads up this is literally the output from yaml linter

joonas:
I think this is a valid change still, not sure if you wanted to try to bring this up to date with the latest chart version so we can maybe get it merged in?

Issue #417, opened 2024-09-16
[FEAT] Allow re-use of existing streams


brooksmtownsend:
Currently, if streams are created per wadm's requirements (e.g. `wadm_events`, `wadm_state`, etc) and their stream parameters don't match, wadm will recreate the stream. This could be something like changing the stream's persistence, replicas, max byte parameters, etc.

This is particularly a problem for Synadia's NGS, because there are account-based limits imposed on streams that wadm will attempt to override.

In the meantime, for deployment scenarios like NGS, exposing more options for the streams may prove to be a viable workaround.

brooksmtownsend:
For some context here, we switched to re-creating the stream here in the past when we changed the stream configuration and we wanted to ensure that we were able to start without issues. There's still a use case for this, but we should approach it on a case-by-case basis and take the burden of checking stream configuration one piece at a time.

brooksmtownsend:
Good news, this already works. Closing this

PR #418, opened 2024-09-17
feat(wadm)!: support configuring max stream bytes

brooksmtownsend:
## Feature or Problem
This PR surfaces additional command line flags & environment variables for configuring the creation of NATS streams and KV buckets with a maximum amount of bytes. By default there is no limit, but depending on the NATS deployment, operators may want to (or need to) supply this parameter.

I hid these variables by default because they are very verbose, but I plan on referencing them in documentation.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -161,6 +161,75 @@ struct Args {
     /// account. See the deployment guide for more information
     #[arg(long = "multitenant", env = "WADM_MULTITENANT", hide = true)]
     multitenant: bool,
+
+    //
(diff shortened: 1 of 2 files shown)


ricochet (approved):

ahmedtadde (approved):

brooksmtownsend:
Ah @ahmedtadde yep! Mostly what instigated this was #417, we had a dev in the wasmCloud slack who wanted to use https://www.synadia.com/ for their NATS infrastructure and you have to supply the max_bytes parameter for creating the streams. It would probably be helpful to have a script that creates these streams based on this configuration so you can pre-create the streams in any NATS infra

PR #421, opened 2024-09-20
chore: Rework release pipeline

### LINKED ISSUE #210: Build a static linux binary
Linux binaries should be compiled towards `musl`, not GNU, such that they are static and portable.

joonas:
## Feature or Problem

We'd like to produce Linux `musl` binaries, so I took the opportunity to consolidate and clean up the release pipeline holistic while switching the Linux builds over from `gnu` to `musl`.

## Related Issues

Fixes #210

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,118 +9,97 @@ on:
       - 'client-v*'
   workflow_dispatch: # Allow manual creation of artifacts without a release
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   build:
     name: build release assets
-    runs-on: ${{ matrix.config.os }}
+    runs-on: ${{ matrix.config.runnerOs }}
     outputs:
       version_output: ${{ steps.version_output.outputs.version }}
     strategy:
       matrix:
         config:
-          # NOTE: We are building on an older version of ubuntu because of libc compatibility
-          # issues. Namely, if we build on a new version of libc, it isn't backwards compatible with
-          # old versions. But if we build on the old version, it is compatible with the newer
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -9,118 +9,97 @@ on:
...
+              target: 'aarch64-unknown-linux-musl',
+              uploadArtifactSuffix: 'linux-aarch64',
+              buildOutputPath: 'target/aarch64-unknown-linux-musl/release/wadm',
             }
           - {
-              os: 'macos-13',
-              arch: 'amd64',
-              extension: '',
-              targetPath: 'target/release/',
+              runnerOs: 'macos-14-large',
```
thomastaylor312, line 38: Is there a reason this one is using large?

thomastaylor312 (commented):
Just one small question on this. Otherwise this is looking good

.github/workflows/release.yml
joonas: I picked these values based on my reading of the [Available Images chart GitHub publishes for their runner images](https://github.com/actions/runner-images?tab=readme-ov-file#available-images), based on my understanding, if you want an arm64-based runner you have to pick `macos-14-large` instead of `macos-14`

joonas: We could also use `macos-13` for x86 and `macos-14` for arm64, but that felt a bit strange, open to suggestions though!

joonas: Ended up using the same runner image for both the x86 and arm64 builds.

thomastaylor312 (approved):

PR #409, opened 2024-09-04
deps(async-nats): upgrade to 0.36

brooksmtownsend:
## Feature or Problem
This PR upgrades to async-nats 0.36, which will stay in draft until wasmcloud-control-interface 1.1.0 is released

PR #431, opened 2024-09-26
chore(deps): update control interface to v2.0.0

vados-cosmonic:
## Feature or Problem

NOTE: this branch won't compile since it's using a local reference to my local control interface v2.0.0, but once we have released `wasmcloud-control-interface` we can change that -- the changed code will be the same

### DIFF
--- a/crates/wadm/src/scaler/spreadscaler/link.rs
+++ b/crates/wadm/src/scaler/spreadscaler/link.rs
@@ -149,12 +149,15 @@ where
                 (
                     true,
                     // TODO(#88): reverse compare too
-                    // Ensure all named configs are the same
-                    linkdef.source_config.iter().all(|config_name| {
-                        self.config.source_config.iter().any(|c| c == config_name)
-                    }) || linkdef.target_config.iter().all(|config_name| {
-                        self.config.target_config.iter().any(|c| c == config_name)
-                    }),
+                    // Ensure all supplied configs (both source and target) are the same
+                    linkdef
+                        .source_config
+                        .iter()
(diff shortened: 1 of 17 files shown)


brooksmtownsend:
#409 depends on this

crates/wadm/src/scaler/spreadscaler/link.rs
```diff
@@ -149,12 +149,15 @@ where
...
-                        self.config.source_config.iter().any(|c| c == config_name)
-                    }) || linkdef.target_config.iter().all(|config_name| {
-                        self.config.target_config.iter().any(|c| c == config_name)
-                    }),
+                    // Ensure all supplied configs are the same
+                    linkdef
+                        .source_config
+                        .iter()
+                        .eq(self.config.source_config.iter())
+                        || linkdef
```
brooksmtownsend, line 157: Is this a bug in the original implementation, should this be a `&&`?

e.g. all source config is the same & all target config is the same?

brooksmtownsend (commented):
Lots o-changes but looks really nice! Pending comment because it's not related to this PR

crates/wadm/src/scaler/spreadscaler/link.rs
vados-cosmonic: Yeah actually I thought it was -- but wasn't sure... Is there any reason you'd want one *or* the other...?

brooksmtownsend: Maybe the original logic was something like `if source_different || target_different`

brooksmtownsend (approved):

Issue #419, opened 2024-09-17
[FEATURE]: Add regex matching to spreadscalers


jamesstocktonj1:
# Description
This feature lets you deploy to a set of labels which match the RegEx rather than just being equal to a label. This could allow for the following:
- may reduce the number of labels per-host
- keep traffic within a single availability zone (for use with Nats Leaf Nodes)
- should be backwards compatible (provided your label doesn't have \, *, + etc...)

## Example
Imagine a set of 5 nodes, each with the following `zone` labels:
- us-east-1
- us-east-2
- us-west-1
- us-west-2
- eu-west-1

And the following spreadscaler config:
```
- type: spreadscaler
  properties:
    replicas: 100
    spread:
      - name: spread-a
        requirements:
          zone: ([a-z]{2})-west-1
```

Would result in deployment to:
- us-west-1
- eu-west-1

This prevents the need for separate labels for country, region and availability zone.

brooksmtownsend:
Ooh @jamesstocktonj1 this is a really cool feature idea! I don't think this would be all that difficult to implement in the spreadscalers + daemonscalers, and we might be able to just treat the whole thing as regex. We should publish recommendations here if there are issues with escaping + special characters in YAML strings.

Notably this would not allow for regex matching on the key of the label, which I think is okay. Happy to help point in the right direction to where this can be implemented

jamesstocktonj1:
@brooksmtownsend Thanks for the feedback! I have used regex within yaml files before and they can just be wrapped in quote marks if/when escape characters become an issue. But I will be sure to check this out when testing.

LiamRandall:
This is a cool idea; I can imagine a lot of different ways this could be leveraged.

PR #433, opened 2024-09-28
feat: regex matching for scaler requirements

jamesstocktonj1:
## Feature or Problem
This features allows for scaler configurations to match a regex expression rather than the exact label value. All spread requirements will now be treated as regex with (most) existing wadm files being backwards compatible in matching the labels. This gives more flexible deployment and can be used in addition to leaf nodes to keep traffic within the same area.

## Related Issues
- #419 

## Consumer Impact
- escape characters may need to be used with hosts which contain regex-specific characters (e.g. *, +, ? ...). For example to match the label `star-*` you would need the regex `star-\*`

### Unit Test(s)
- existing unit tests remain untouched to ensure backwards compatability
- additional unit tests for both component and provider
- additional e2e test for regex matching
- additional serialisation and deserialisation tests

### Manual Verification
- various wadm files tested on a simple 4 host setup

### DIFF
--- a/crates/wadm-types/src/lib.rs
+++ b/crates/wadm-types/src/lib.rs
@@ -722,6 +722,12 @@ mod test {
             weight: Some(20),
         };
         spread_vec.push(spread_item);
+        let spread_item = Spread {
+            name: "northcoast".to_string(),
+            requirements: BTreeMap::from([("zone".to_string(), "([a-z]{2}-north-([0-9])".to_string())]),
+            weight: Some(40),
+        };
+        spread_vec.push(spread_item);
         let mut trait_vec: Vec<Trait> = Vec::new();
         let spreadscalerprop = SpreadScalerProperty {
             instances: 4,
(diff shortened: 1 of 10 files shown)


brooksmtownsend:
Hey @jamesstocktonj1 ! I've been chatting it over with some of the other wadm maintainers, and I think there's a general concern with Regex being implemented as-is here today and how long we could support that feature. I wanted to give you a heads up that we're likely going to ask for a bit of a refactor in this PR so that the Regex functionality can fit more into a model like the [Nomad constraint](https://developer.hashicorp.com/nomad/docs/job-specification/constraint) feature, for example:
```yaml
        - type: spreadscaler
          properties:
            instances: 1
            constraints:
              - operation: "=="
                label: "foo"
                value: "bar"
              - operation: "regex"
                attribute: "${labels.foo}"
                value: ".*"
```

This would be a holistic improvement to our spread logic and isn't a fault of your PR. I didn't want to request changes just yet because the maintainers need to get together and decide on the right structure before we ask you to change anything, but I wanted to give you an update. Do you have any thoughts/concerns/questions for right now?

Thank you so much for taking the initiative to implement this + add tests to verify that it works, this is really a fantastic contribution.

PR #455, opened 2024-10-15
chore: Pull in slightly older version of regex

thomastaylor312:
Because transitive deps suck. We need this so we can update the OCI deps in the main host


joonas (approved):

brooksmtownsend (approved):
fun, need to yank the `*.*.0` versions then after this releases

thomastaylor312:
Yep, yanking shall occur

PR #467, opened 2024-10-24
fix(chart): reference the correct value for the Jetstream domain

protochron:
## Feature or Problem
Pull the `jetstreamDomain` value from `config.wadm.nats.jetstreamDomain` instead of from `config.wadm.jetstreamDomain` since that is what we define in the values file. It is also a more logical way to group the value than what the chart was expecting.

## Release Information
Next chart release

### DIFF
--- a/charts/wadm/Chart.yaml
+++ b/charts/wadm/Chart.yaml
@@ -15,10 +15,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: "0.2.7"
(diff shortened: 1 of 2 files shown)


charts/wadm/Chart.yaml
```diff
@@ -15,10 +15,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: "0.2.7"
+version: "0.2.8"
```
joonas, line 18: I would say you can even cut a `0.3.0` for this if you'd like, I'm fine either way

joonas (approved):

Issue #432, opened 2024-09-28
[BUG] Infinite reconciliation loop when overlapping rules exist


jamesstocktonj1:
# Description

Consider a scenario with 4 wasmcloud instances with the following labels:
- node 1 - region=us-east-1, cloud=real
- node 2 - region=us-east-2, cloud=real
- node 3 - region=us-west-1, cloud=real
- node 4 - region=us-west-2, cloud=real

And the following spreadscale properties for a component:
```
type: spreadscaler
properties:
  instances: 12
  spread:
    - name: eastcoast
      requirements:
        region: us-east-1
      weight: 25
    - name: westcoast
      requirements:
        region: us-west-1
      weight: 25
    - name: realcloud
      requirements:
        cloud: real
      weight: 50
```

## Problem
At first, wadm deploys 3 instances to `node 1` and 3 instances to `node 3`. This satisfies the requirements `eastcoast` and `westcoast` but it also satisfies the requirement `realcloud`. Even though all the requirements are satisfied, the deployment is not successful because the total number of components deployed is 6 and not 12. Therefore Wadm continues the reconciliation process.

After this, in order to try and satisfy the number of instances it deploys 6 instances on either `node 2` or `node 4`. This then satisfies the total number of instances but does not satisfy the weighting for `realcloud` as it is 100% and not 50%.

This cycle continues as wadm attempts to satisfy all conditions for deployment, resulting in a reconciliation loop with no exit.

## TLDR
A spreadscaler config may have a set of requirements which cannot fully be satisfied so the deployment status remains as "Reconciling".

brooksmtownsend:
@ahmedtadde fixed this in #491

PR #491, opened 2024-11-13
fix: detect spread scaler requirements violation

ahmedtadde:
## Feature or Problem
A `ComponentSpreadScaler` config may have a set of requirements which cannot fully be satisfied so the deployment status remains as "Reconciling".

## Related Issues
#432 

## Release Information
`next`

## Consumer Impact
This is an enhancement/correctness change. It should have no adverse effect on existing functionality. This change prevents potential `Heisenbugs` where `wadm` could potentially operate on faulty instance spread without any detection.

## Testing
All existing tests (unit and e2e) are functional

### Unit Test(s)
Added two test cases `can_detect_spread_requirements_violation_1` and `can_detect_spread_requirements_violation_2`

### DIFF
--- a/crates/wadm/src/scaler/spreadscaler/mod.rs
+++ b/crates/wadm/src/scaler/spreadscaler/mod.rs
@@ -1,5 +1,6 @@
-use std::collections::{BTreeMap, HashSet};
-use std::{cmp::Ordering, cmp::Reverse, collections::HashMap};
+use std::{
+    cmp::Ordering, cmp::Reverse, collections::BTreeMap, collections::HashMap, collections::HashSet,
+};
 
 use anyhow::Result;
 use async_trait::async_trait;
@@ -171,6 +172,7 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ComponentSpreadScaler<S> {
 
         let mut spread_status = vec![];
         trace!(spread_requirements = ?self.spread_requirements, ?component_id, "Computing commands");
+        let mut component_instances_per_eligible_host: HashMap<&String, usize> = HashMap::new();
         let commands = self
             .spread_requirements
             .iter()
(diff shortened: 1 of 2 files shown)


crates/wadm/src/scaler/spreadscaler/mod.rs
```diff
@@ -470,6 +490,94 @@ fn compute_spread(spread_config: &SpreadScalerProperty) -> Vec<(Spread, usize)>
...
+
+        spread_instances.insert(spread.name.clone(), (projected_count, *target_count));
+    }
+
+    // Step 4: Compare the tuples' values to detect conflicts
+    let mut violations = Vec::new();
+    for (spread_name, (projected_count, target_count)) in spread_instances {
+        if projected_count != target_count {
+            violations.push(format!(
+                "Spread requirement violation: {} spread requires {} instances vs {} computed from reconcialiation commands",
```
brooksmtownsend, line 568:
```suggestion
                "Spread requirement violation: {} spread requires {} instances vs {} computed from reconciliation commands", 
```

spelling nit

brooksmtownsend: Regarding the status message, what do you think about rewording "violation" to "spread requirement conflict", indicating that there are conflicting spreads that are messing with the reconciliation? Trying to think of the end user when looking at that status message, they should have an idea of what the problem is and I think noting that there are conflicts might help

crates/wadm/src/scaler/spreadscaler/mod.rs
```diff
@@ -266,6 +276,16 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ComponentSpreadScaler<S> {
...
 
+        // Detect spread requirement violations
+        if let Some(message) = detect_spread_requirement_violations(
+            &self.spread_requirements,
+            &hosts,
+            &component_instances_per_eligible_host,
+            &commands,
+        ) {
+            spread_status.push(StatusInfo::failed(&message));
+        }
```
brooksmtownsend, line 287: In this case, where there are conflicting spreads, doesn't that mean that the scalers will reconcile forever trying to adjust the number of components on a host? I might be misunderstanding, but if that's the case, what do you think about returning no commands here in order to ensure that we don't deadloop?

brooksmtownsend (commented):
Logic looks great, and tests are 👨🏻‍🍳 💋. Just a comment around the wording of the status message and thoughts around reconciliation

brooksmtownsend (approved):
Great stuff @ahmedtadde

Issue #445, opened 2024-10-08
[FEAT] Validate manifests to ensure links are unique


brooksmtownsend:
As of wasmCloud v1.3.0, you cannot define links from the same source on the same namespace, package, and interface but to two different targets _without_ specifying a separate link name. This technically wasn't checked before in wasmCloud but it never worked for wasmCloud components, and only worked for capability providers because they are able to synthesize additional information in their implementations.

We need a validation rule to ensure that this is not possible when deploying a manifest, since it's just frustrating to have this pop up as an issue afterwards. In this validation rule, the following manifest should fail:
```yaml
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: test-link-name-uniqueness
  annotations:
    description: 'test'
spec:
  components:
    - name: http-component
      type: component
      properties:
        image: file://./build/http_hello_world_s.wasm
      traits:
        - type: spreadscaler
          properties:
            instances: 1
    - name: http-component-two
      type: component
      properties:
        image: file://./build/http_hello_world_s.wasm
      traits:
        - type: spreadscaler
          properties:
            instances: 1
    - name: httpserver
      type: capability
      properties:
        image: ghcr.io/wasmcloud/http-server:0.22.0
      traits:
        - type: link
          properties:
            target: http-component
            namespace: wasi
            package: http
            interfaces: [incoming-handler]
            source_config:
              - name: default-http
                properties:
                  address: 127.0.0.1:8080
        - type: link
          properties:
            target: http-component-two
            namespace: wasi
            package: http
            interfaces: [incoming-handler]
            source_config:
              - name: default-http-two
                properties:
                  address: 127.0.0.1:8081
```

Note that above the links are effectively identical, but with different targets and configuration. The correct traits block is as follows:
```yaml
      traits:
        - type: link
          properties:
            target: http-component
            name: default # for clarity, specifying "default" is the same as not specifying
            namespace: wasi
            package: http
            interfaces: [incoming-handler]
            source_config:
              - name: default-http
                properties:
                  address: 127.0.0.1:8080
        - type: link
          properties:
            target: http-component-two
            name: link-two # can be named anything, but must not be "default"
            namespace: wasi
            package: http
            interfaces: [incoming-handler]
            source_config:
              - name: default-http-two
                properties:
                  address: 127.0.0.1:8081
```

Vikrantpalle:
Hey, can I pick this issue up?

[...]

PR #507, opened 2024-12-08
Add manifest validation check for "duplicate" links

Vikrantpalle:
## Feature or Problem

Added a manifest validation check for "duplicate" links.

Multiple links from the same source with the same name, namespace and package are considered duplicate links.

## Related Issues

- #445 

## Consumer Impact

- Manifests that previously passed validation checks will not pass validation checks anymore if they contain duplicate links

### Unit Test(s)

- Modified the manifest validation tests to include a duplicate links test

### Manual Verification

- verified the output for possible cases

### DIFF
--- a/crates/wadm-types/src/validation.rs
+++ b/crates/wadm-types/src/validation.rs
@@ -729,8 +729,8 @@ pub fn is_valid_label_name(name: &str) -> bool {
 /// are considered duplicate links.
 fn check_duplicate_links(manifest: &Manifest) -> Vec<ValidationFailure> {
     let mut failures = Vec::new();
-    let mut link_ids = HashSet::new();
     for component in manifest.components() {
+        let mut link_ids = HashSet::new();
         for link in component.links() {
             if let TraitProperty::Link(LinkProperty {
                 name,
@@ -740,7 +740,6 @@ fn check_duplicate_links(manifest: &Manifest) -> Vec<ValidationFailure> {
             }) = &link.properties
             {
                 if !link_ids.insert((
-                    component.name.clone(),
                     name.clone()
                         .unwrap_or_else(|| DEFAULT_LINK_NAME.to_string()),
                     namespace,
--- a/crates/wadm-types/src/validation.rs
+++ b/crates/wadm-types/src/validation.rs
@@ -11,7 +11,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     CapabilityProperties, ComponentProperties, LinkProperty, Manifest, Properties, Trait,
-    TraitProperty, LATEST_VERSION,
+    TraitProperty, DEFAULT_LINK_NAME, LATEST_VERSION,
 };
 
 /// A namespace -> package -> interface lookup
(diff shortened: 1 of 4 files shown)


crates/wadm-types/src/validation.rs
```diff
@@ -722,6 +723,50 @@ pub fn is_valid_label_name(name: &str) -> bool {
...
+                            name.clone()
+                                .unwrap_or_else(|| DEFAULT_LINK_NAME.to_string()),
+                            namespace,
+                            package,
+                            interface,
+                        )) {
+                            failures.push(ValidationFailure::new(
+                                ValidationFailureLevel::Error,
+                                format!(
+                                    "Duplicate links found inside component: {}",
```
joonas, line 757: I wonder if the error message should contain information about which link name (or perhaps namespace/package/interface triplet) conflicted so that it might be easier to track down?

Vikrantpalle:
I removed interfaces when searching for duplicate links as it doesn't look like it's being checked in https://github.com/wasmCloud/wasmCloud/pull/2963 either.

crates/wadm-types/src/validation.rs
Vikrantpalle: I guess both the link name and the namespace/package can be included in the message?

crates/wadm-types/src/validation.rs
```diff
@@ -722,6 +723,47 @@ pub fn is_valid_label_name(name: &str) -> bool {
...
+    let mut link_ids = HashSet::new();
+    for component in manifest.components() {
+        for link in component.links() {
+            if let TraitProperty::Link(LinkProperty {
+                name,
+                namespace,
+                package,
+                ..
+            }) = &link.properties
+            {
```
brooksmtownsend, line 741:
```suggestion
    for component in manifest.components() {
        let mut link_ids = HashSet::new();
        for link in component.links() {
            if let TraitProperty::Link(LinkProperty {
                name,
                namespace,
                package,
                ..
            }) = &link.properties
            {
```

nit: if we move the declaration of the `link_ids` here inside of the loop, it would reset on each iteration and you could omit the `component.name.clone()` when inserting into the hashset.

crates/wadm/src/server/handlers.rs
```diff
@@ -1107,6 +1107,16 @@ mod test {
...
+
+        let manifest = deserialize_yaml("../../tests/fixtures/manifests/duplicate_links.yaml")
+            .expect("Should be able to parse");
+
+        match validate_manifest(&manifest).await {
+            Ok(()) => panic!("Should have detected duplicate links"),
+            Err(e) => assert!(e
+                .to_string()
+                .contains("Duplicate link found inside component")),
+        }
```
brooksmtownsend, line 1119:
```suggestion
        assert!(e.is_err_and(|e| e.to_string().contains("Duplicate link found inside component"));
```

This could be simplified a bit with https://doc.rust-lang.org/std/result/enum.Result.html#method.is_err_and

brooksmtownsend (approved):
I only have two small code style nits, neither of them should block this from going in 😄 @Vikrantpalle please feel free to take a look and if you'd like to make the change it will be welcome, if you think it hurts readability/understanding of the functionality just resolve the comment and we can merge

crates/wadm/src/server/handlers.rs
Vikrantpalle: Since the other sub-tests in this function follow a similar syntax I think I'll leave it as is.

brooksmtownsend (approved):
Great work @Vikrantpalle ! Thank you for fixing this

PR #513, opened 2024-12-10
fix(server): wash app delete command returns success for invalid application name

### LINKED ISSUE #502: [BUG] wash app delete command returns success for invalid application name
## Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [x] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

## Describe the bug

If an incorrect model name is passed without a version to the `wash app del <app_name>` command, it still outputs a success message ("Deleted application: <app_name>") even though this app does not exist.

## Steps to reproduce

- list all applications with `wash app list`
- try to delete an app using an invalid name without specifying a version
- check the apps again with `wash app list` to confirm no changes have occurred

Example:

```sh
$ wash app list
  Name                                  Deployed Version                      Status
  rust-hello-world                      01JE99W346KMJR9XG33FE7QBYM            Deployed
    └ HTTP hello world demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)
```

```sh
$ wash app del test
Deleted application: test
```

```sh
$ wash app list
  Name                                  Deployed Version                      Status
  rust-hello-world                      01JE99W346KMJR9XG33FE7QBYM            Deployed
    └ HTTP hello world demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)
```

## Expected behavior

The `wash app del` command should return a meaningful message for invalid app names instead of printing a success message. The output should be the same as for invalid versions:

```sh
$ wash app del test 1.0.0
Already deleted application: test
```

## Environment

- OS: MacOS
- Shell: zsh
- wasmcloud Versions:
  ```sh
  $ wash -V
  wash          v0.37.0
  ├ nats-server v2.10.20
  ├ wadm        v0.18.0
  └ wasmcloud   v1.4.2
  ```

## Root Cause

`wadm` only checks wether the application actually exists if a version has been specified (see [here](https://github.com/wasmCloud/wadm/blob/main/crates/wadm/src/server/handlers.rs#L417)). If no version was specified, `wadm` simply deletes the application manifest (see [here](https://github.com/wasmCloud/wadm/blob/main/crates/wadm/src/server/handlers.rs#L493)) without covering the `DeleteResult::Noop` case.

## Proposal

Checking first if for the specified application a manifest exists and start a deletion attempt only in case it does. Otherwise return early with `DeleteResult::Noop`.

If you agree with my proposal, I would be happy to create a PR for it.

ffuerste:
## Feature or Problem

If an incorrect model name is passed without a version to the `wash app del <app_name>` command, it still outputs a success message ("Deleted application: <app_name>") even though this app does not exist.

## Related Issues

closes #502 

## Release Information

This PR fixes a non-critical issue. It can therefore be safely included in the next version.

## Consumer Impact

This PR adapts the lookup logic for deletion requests, so that also for non-existing models a `DeleteResult::Noop` result is returned if no version is specified. This changes the current return value (which is `DeleteResult::Deleted`).

Currently, the documented and recommended way to delete applications is to use the wadm-client crate. The wadm-client create [interprets both delete results as a success](https://docs.rs/wadm-client/latest/wadm_client/struct.Client.html#method.delete_manifest):

```rust
        match body.result {
            DeleteResult::Error => Err(ClientError::ApiError(body.message)),
            DeleteResult::Noop => Ok(false),
            DeleteResult::Deleted => Ok(true),
        }
```

The `wash` CLI uses the wadm-client crate internally and outputs based on the contained result value (`true` or `false`) different output messages.

### Unit Test(s)

No unit tests have been added or updated.

### Acceptance or Integration

An additional test case was added to the `test_delete_noop` integration test (covering the deletion of a non-existing model without specifying a version).

### DIFF
--- a/crates/wadm/src/server/handlers.rs
+++ b/crates/wadm/src/server/handlers.rs
@@ -412,98 +412,110 @@ impl<P: Publisher> Handler<P> {
                 }
             }
         };
+
         // TODO(#451): if shared and deployed, make sure that no other shared apps are using it
-        let reply_data = if let Some(version) = req.version {
+        let reply_data = {
             match self.store.get(account_id, lattice_id, name).await {
                 Ok(Some((mut current, current_revision))) => {
-                    let deleted = current.delete_version(&version);
-                    if deleted && !current.is_empty() {
-                        // If the version we deleted was the deployed one, undeploy it
-                        let deployed_version = current.deployed_version();
-                        let undeploy = if deployed_version.map(|v| v == version).unwrap_or(false) {
-                            trace!(?deployed_version, deleted_version = %version, "Deployed version matches deleted. Will undeploy");
-                            current.undeploy();
-                            true
+                    if let Some(version) = req.version {
+                        let deleted = current.delete_version(&version);
+                        if deleted && !current.is_empty() {
+                            // If the version we deleted was the deployed one, undeploy it
+                            let deployed_version = current.deployed_version();
+                            let undeploy = if deployed_version
+                                .map(|v| v == version)
+                                .unwrap_or(false)
+                            {
+                                trace!(?deployed_version, deleted_version = %version, "Deployed version matches deleted. Will undeploy");
+                                current.undeploy();
+                                true
+                            } else {
+                                trace!(?deployed_version, deleted_version = %version, "Deployed version does not match deleted version. Will not undeploy");
+                                false
+                            };
+                            self.store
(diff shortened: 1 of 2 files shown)


thomastaylor312 (approved):
Excellent work here @ffuerste. Thanks for the fix!

PR #517, opened 2024-12-14
feat: Adds flake

thomastaylor312:
This adds a `flake.nix` and associated files for a dev environment via nix and steps for building all the parts of wadm in an incremental way. Part of this was an experiment to see what we could get nix to do with monorepo deps in the simple case. Please note that this does not actually add this as part of the build pipeline, just adds the file the repo for people to use. Maybe in the future we can add in builds and caching into our pipelines.


thomastaylor312:
Special thanks to @rvolosatovs for helping me figure out the cargo.toml munging

tests/api_model_operations.rs
```diff
@@ -472,12 +472,10 @@ async fn test_delete_noop() {
     let resp: DeleteModelResponse = test_server
         .get_response(
             "default.model.del.my-example-app",
-            serde_json::to_vec(&DeleteModelRequest {
-                version: None,
-            }).unwrap(),
+            serde_json::to_vec(&DeleteModelRequest { version: None }).unwrap(),
```
thomastaylor312, line 475: Note: This changes were just a missing `cargo fmt` that I found while running tests in the flake

flake.nix
```diff
@@ -0,0 +1,270 @@
...
+        # Common arguments can be set here to avoid repeating them later
+        commonArgs = {
+          inherit src;
+          strictDeps = true;
+
+          buildInputs = [
+            # Add additional build inputs here
+          ] ++ lib.optionals pkgs.stdenv.isDarwin [
+            # Additional darwin specific inputs can be set here
+            pkgs.libiconv
```
rvolosatovs, line 55: Is this actually needed in 24.11?

flake.nix
```diff
@@ -0,0 +1,270 @@
...
+          } ''
+            touch $out
+          '';
+        };
+
+        packages = {
+          inherit wadm wadm-client wadm-types wadm-lib;
+          default = wadm;
+        } // lib.optionalAttrs (!pkgs.stdenv.isDarwin) {
+          workspace-llvm-coverage = craneLibLLvmTools.cargoLlvmCov
```
rvolosatovs, line 244: Is this not supported on Darwin?

rvolosatovs (approved):

flake.nix
thomastaylor312: I copied this from the crane docs, so I am not sure if it isn't or isn't needed

flake.nix
thomastaylor312: I don't actually know! This was in their docs, but I can try without it

thomastaylor312: Yep, wasn't needed on 24.11

Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 4
+version = 3
```
brooksmtownsend, line 3: is this a symptom of running `nix build` rather than `cargo build` here?

brooksmtownsend (approved):
All looks good to me, only a supernit and a question. Is it worth throwing a comment at the top of the flake marking it as experimental? Only to consider whether or not we're going to update & maintain it often

Cargo.lock
thomastaylor312: Yeah, I just deleted it and retried again and it still is version 3. I'm on rust 1.83, so it should be correct

PR #518, opened 2024-12-14
allow unique interfaces across duplicate links and test

LUK3ARK:
Related fix for https://github.com/wasmCloud/wasmCloud/issues/3801


LUK3ARK:
I didn't see any changes in WADM required, seems that it treats each interface explicitly anyway. Let me know if I have missed anything other than this validation level.

brooksmtownsend (changes requested):
This all looks great, we should just hold it back until wasmCloud supports this same matching validation

LUK3ARK:
What are we still waiting for with this? Major blocker for me at the moment

brooksmtownsend (approved):

Issue #447, opened 2024-10-09
[FEAT] Reflect unhealthy providers in application status


brooksmtownsend:
The Provider spreadscaler + daemonscalers currently handle `provider_start_failed` events to update application status based on application level events. We should extend this handling to also include `health_check_failed` events and `health_check_succeeded` in order to reflect when an application is unhealthy.

We could start this feature using the existing `Failed` status to reflect the issue, but we may want to add an `Unhealthy` status to indicate that the application was scheduled/deployed but with problems.

See https://wasmcloud.com/docs/reference/cloud-event-list#health_check_passed for the event specifications.

Vikrantpalle:
Hey, I would like to have a go at this issue.

PR #524, opened 2024-12-17
Reflect unhealthy providers in application status

Vikrantpalle:
## Feature or Problem

Scaler status returns a `Failed` status if it contains any providers which have failed their most recent health check and are in a `Failed` state.

## Related Issues

- #447 

### Unit Test(s)

Added 2 unit tests each for daemon and spread scalers
- test_unhealthy_providers_return_unhealthy_status
- test_healthy_providers_return_healthy_status
### Manual Verification

Verified the scaler status after `reconcile` is run and that all unit tests passed.

### DIFF
--- a/crates/wadm/src/scaler/daemonscaler/provider.rs
+++ b/crates/wadm/src/scaler/daemonscaler/provider.rs
@@ -122,15 +122,31 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ProviderDaemonScaler<S> {
                 });
                 let status = self.status.read().await.to_owned();
                 // update health status of scaler
-                if let Some(status) = match (status.status_type, unhealthy_providers > 0) {
+                if let Some(status) = match (status, unhealthy_providers > 0) {
                     // scaler is deployed but contains unhealthy providers
-                    (StatusType::Deployed, true) => Some(StatusInfo::unhealthy(&format!(
+                    (
+                        StatusInfo {
+                            status_type: StatusType::Deployed,
+                            ..
+                        },
+                        true,
+                    ) => Some(StatusInfo::failed(&format!(
                         "Unhealthy provider on {} host(s)",
                         unhealthy_providers
                     ))),
                     // scaler can become unhealthy only if it was previously deployed
                     // once scaler becomes healthy again revert back to deployed state
-                    (StatusType::Unhealthy, false) => Some(StatusInfo::deployed("")),
+                    // this is a workaround to detect unhealthy status until
+                    // StatusType::Unhealthy can be used
+                    (
+                        StatusInfo {
+                            status_type: StatusType::Failed,
+                            message,
+                        },
+                        false,
+                    ) if message.starts_with("Unhealthy provider on") => {
+                        Some(StatusInfo::deployed(""))
+                    }
                     // don't update status if scaler is not deployed
                     _ => None,
                 } {
@@ -785,7 +801,7 @@ mod test {
 
         assert_eq!(
             spreadscaler.status.read().await.to_owned(),
-            StatusInfo::unhealthy("Unhealthy provider on 1 host(s)")
+            StatusInfo::failed("Unhealthy provider on 1 host(s)")
         );
         Ok(())
     }
--- a/crates/wadm/src/scaler/daemonscaler/provider.rs
+++ b/crates/wadm/src/scaler/daemonscaler/provider.rs
@@ -4,6 +4,7 @@ use anyhow::Result;
 use async_trait::async_trait;
 use tokio::sync::RwLock;
 use tracing::{instrument, trace};
+use wadm_types::api::StatusType;
 use wadm_types::{api::StatusInfo, Spread, SpreadScalerProperty, TraitProperty};
 
 use crate::commands::StopProvider;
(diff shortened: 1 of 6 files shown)


crates/wadm/src/scaler/daemonscaler/provider.rs
```diff
@@ -81,10 +85,12 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ProviderDaemonScaler<S> {
...
             | Event::ProviderStopped(ProviderStopped { provider_id, .. })
-                if provider_id == &self.config.provider_id =>
-            {
-                self.reconcile().await
-            }
+            | Event::ProviderHealthCheckFailed(ProviderHealthCheckFailed {
+                data: ProviderHealthCheckInfo { provider_id, .. },
+            })
+            | Event::ProviderHealthCheckPassed(ProviderHealthCheckPassed {
+                data: ProviderHealthCheckInfo { provider_id, .. },
```
brooksmtownsend, line 92: In this case, I actually think that we should update the status in the case of a failed health check but not run a full reconcile. The main reason being that for a failed health check we don't really want to try and reconcile with commands, but we do want to update the status to be Failed. Does that make sense?

brooksmtownsend: Your other modifications in `reconcile()` look good though, and I think that would be able to show the failed status after we set it in response to that event. Here (if I remember right) there's also a message on the event that you can set the status to

crates/wadm/src/scaler/spreadscaler/provider.rs
```diff
@@ -99,10 +100,12 @@ impl<S: ReadStore + Send + Sync + Clone> Scaler for ProviderSpreadScaler<S> {
...
-                if provider_id == &self.config.provider_id =>
-            {
-                self.reconcile().await
-            }
+            | Event::ProviderHealthCheckFailed(ProviderHealthCheckFailed {
+                data: ProviderHealthCheckInfo { provider_id, .. },
+            })
+            | Event::ProviderHealthCheckPassed(ProviderHealthCheckPassed {
+                data: ProviderHealthCheckInfo { provider_id, .. },
+            }) if provider_id == &self.config.provider_id => self.reconcile().await,
```
brooksmtownsend, line 108: Same comment here re: setting status rather than reconciling

crates/wadm/src/scaler/spreadscaler/provider.rs
```diff
@@ -1206,4 +1229,226 @@ mod test {
...
+        store
+            .store(
+                lattice_id,
+                host_id_one.to_string(),
+                Host {
+                    components: HashMap::new(),
+                    friendly_name: "hey".to_string(),
+                    labels: HashMap::from_iter([
+                        ("cloud".to_string(), "fake".to_string()),
+                        ("region".to_string(), "us-noneofyourbusiness-1".to_string()),
```
brooksmtownsend, line 1253: 😆

crates/wadm/src/scaler/spreadscaler/provider.rs
```diff
@@ -1206,4 +1229,226 @@ mod test {
...
+                },
+            )
+            .await?;
+
+        spreadscaler.reconcile().await?;
+
+        assert_eq!(
+            spreadscaler.status.read().await.to_owned(),
+            StatusInfo::failed("Unhealthy provider on 1 host(s)")
+        );
```
brooksmtownsend, line 1451: Once the changes are made in the scaler, would love to see the spreadscaler react to the health check failed here 😄

brooksmtownsend (changes requested):
Great work here @Vikrantpalle ! I have one request around how we're handling the health check event here, I think we should update the status with the message on the event but not actually do a full reconcile. Let me know if there's any more detail I can add there, and thank you for adding tests!

crates/wadm/src/scaler/daemonscaler/provider.rs
Vikrantpalle: Yep, makes sense! I'll make the change to set the status without running a full reconcile.

Vikrantpalle: The health check event doesn't have any message attached to it so I haven't made any changes to the status message.

Vikrantpalle:
I added a new `Unhealthy` status because I thought it would be more clear and wouldn't require matching the message inside the failed status to figure out whether a scaler is unhealthy. Feel free to let me know if you think this isn't required, I can revert to use the `Failed` status like before.

@brooksmtownsend

ahmedtadde (approved):

brooksmtownsend (changes requested):
Hey @Vikrantpalle I do like the idea of the Unhealthy status, for now though I think we have to use `StatusType::Failed` since using the new status will fail to deserialize on clients where that status doesn't exist yet 😕 It's not ideal, we should probably have a fallback there, but `wash` doesn't ATM.

I think it's good to keep the `Unhealthy` status, but if you can just make sure that the scaler uses `Failed` for now we can release this and then update the scaler in a future version to use `Unhealthy`

brooksmtownsend (approved):
Looks great!

Issue #479, opened 2024-11-08
[FEAT] Add flag for changing stream persistence to `Memory`


brooksmtownsend:
It would be great to be able to configure all of the streams that `wadm` creates to use https://docs.rs/async-nats/0.37.0/async_nats/jetstream/stream/enum.StorageType.html#variant.Memory instead of File, which would be great to create a more ephemeral environment in NATS.

It would be great to be able to have a flag `--stream-persistence` that defaults to `File` but can be overridden to be `Memory`.

markkovari:
Hey @brooksmtownsend I would pick it up this one. Or rather will create a PR, so you can assign it to me. I'll have some questions about some testing, but will ask there rather.

brooksmtownsend:
Fixed in #525

PR #525, opened 2024-12-17
feat(nats): flag to change persistence of streams

markkovari:
## Feature or Problem

cli toggle to modify the nats stream persistence between file and memory.

I do seek some guidance regarding testing here. Since we have multiple tests which have the file - default - mode enabled, should I create another suite or how do you think should I add tests for this?

Thank you

## Related Issues

#479

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -183,13 +183,11 @@ struct Args {
         hide = true
     )]
     max_manifest_bucket_bytes: i64,
-    /// The storage type of nats streams, maps to [`StorageType`](async_nats::jetstream::stream::StorageType)
-    /// Possible values: "File" or "Memory", default is `File`
+    /// The storage type of nats streams, possible values: "file" or "memory", default is "file"
     #[arg(
         long = "stream-persistence",
         env = "WADM_STREAM_PERSISTENCE",
-        default_value_t = StreamPersistence::File,
-        hide = true
+        default_value_t = StreamPersistence::File
     )]
     stream_persistence: StreamPersistence,
     /// Maximum bytes to keep for the command stream
--- a/src/main.rs
+++ b/src/main.rs
@@ -183,7 +183,7 @@ struct Args {
         hide = true
     )]
     max_manifest_bucket_bytes: i64,
-    /// The storage type of nats streams
+    /// Nats streams storage type
     #[arg(
         long = "stream-persistence",
         env = "WADM_STREAM_PERSISTENCE",
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,7 +4,7 @@ use std::time::Duration;
 
 use async_nats::jetstream::{stream::Stream, Context};
 use clap::Parser;
-use nats::StreamPersistance;
+use nats::StreamPersistence;
 use tokio::sync::Semaphore;
 use tracing::log::debug;
 use wadm_types::api::DEFAULT_WADM_TOPIC_PREFIX;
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -182,6 +183,15 @@ struct Args {
         hide = true
     )]
     max_manifest_bucket_bytes: i64,
+    /// The storage type of nats streams, maps to [`StorageType`](async_nats::jetstream::stream::StorageType)
+    /// Possible values: "File" or "Memory", default is `File`
+    #[arg(
+        long = "stream-persistence",
+        env = "WADM_STREAM_PERSISTENCE",
+        default_value_t = StreamPersistence::File,
```
ahmedtadde, line 191: why use `nats::StreamPersistence` as a proxy instead of directly using the mentioned type `[`StorageType`](async_nats::jetstream::stream::StorageType`?

src/main.rs
```diff
@@ -182,6 +183,15 @@ struct Args {
...
+        hide = true
+    )]
+    stream_persistance: StreamPersistence,
```
ahmedtadde, line 194: spelling issue, `stream_persistence`

ahmedtadde:
hey @markkovari thank you for this submission! Upon a first look, I only found a spelling issue that needs to be addressed. Otherwise, the changes made should do the trick.

testing wise, I believe we want the following

- ensure that the flag (parsing) works... as in, when no specified -> expected default value & when specified -> maps to expected values (`File, Memory`)
- ensure all the existing tests work as expected (no regression bugs)... this is basically just running the test suite locally
- To check that we have a functional `Memory` mode, ensure that there is a file when using `File` mode and the absence of said file otherwise (ie, `Memory` mode). This entails figuring out how to find/peek/inspect a JetStream persistence file for a given stream.

src/nats.rs
```diff
@@ -14,6 +14,32 @@ use async_nats::{
 use tracing::{debug, warn};
 use wadm::DEFAULT_EXPIRY_TIME;
 
+#[derive(Debug, Clone, Copy, Default, clap::ValueEnum)]
+#[clap(rename_all = "PascalCase")]
+pub enum StreamPersistence {
+    #[default]
+    File,
+    Memory,
+}
```
brooksmtownsend, line 23: for these, what do you think about making these lowercase instead of pascal case? I know the NATS options are PascalCase, but it feels a little odd to me thinking about doing `wadm --stream-persistence File` over `wadm --stream-persistence file`

src/main.rs
```diff
@@ -182,6 +183,15 @@ struct Args {
         hide = true
     )]
     max_manifest_bucket_bytes: i64,
+    /// The storage type of nats streams, maps to [`StorageType`](async_nats::jetstream::stream::StorageType)
+    /// Possible values: "File" or "Memory", default is `File`
+    #[arg(
+        long = "stream-persistence",
+        env = "WADM_STREAM_PERSISTENCE",
+        default_value_t = StreamPersistence::File,
+        hide = true
```
brooksmtownsend, line 192: I think we should keep this un-hidden since it's a useful flag! (apologies if I said it should be hidden in the issue)

brooksmtownsend (approved):
@markkovari this looks good! If there's any test to run, I think there's just two that we can do:
1. Change one of the e2e tests to use memory backed streams, make sure nothing goes awry there
2. Manually verify that you can run `./wadm --stream-persistence file`, close the wadm instance, then `./wadm --stream-persistence memory` without issue.

brooksmtownsend (approved):

PR #531, opened 2025-01-02
feat(deps): add separate wit-wasm and wit-std features

LUK3ARK:
Up until now wadm-types has only been depended on from a std context, however it would be useful to use it in wasm targets

Change syncs latest wadm-types and allows them to be used in both wasm and std contexts

I couldn't really think of a cleaner way to do the feature gates, open to suggestions


LUK3ARK:
as a note - with this current setup it is currently only possible to depend in a wasm context when the wadm-types dependency has default-features = false

otherwise, tokio causes problems

LUK3ARK:
default-features flag no longer needed for wasm-target compilation

brooksmtownsend (approved):
I'm good with this as long as it works with and without the feature

brooksmtownsend:
tested manually, all LGTM

PR #536, opened 2025-01-02
fix(types): warn on unknown interface

brooksmtownsend:
## Feature or Problem
This PR downgrades the unrecognized interface error into a warning as inspired by some of the recent additions to keyvalue like `batch` and `watch`. We want to maintain a balance in validation to make sure that users don't accidentally link on, say, `wasi:keyvalue/atormics`, but we don't completely block the addition of new interfaces e.g. in wasmcloud:messaging@v0.3.0

Open to thoughts here.

### DIFF
--- a/crates/wadm-types/src/validation.rs
+++ b/crates/wadm-types/src/validation.rs
@@ -160,9 +160,10 @@ fn is_invalid_known_interface(
     };
     // Unknown interface inside known namespace and package is probably a bug
     if !iface_lookup.contains_key(interface) {
-        // Unknown package inside a known interface we control is probably a bug
+        // Unknown package inside a known interface we control is probably a bug, but may be
+        // a new interface we don't know about yet
         return vec![ValidationFailure::new(
-            ValidationFailureLevel::Error,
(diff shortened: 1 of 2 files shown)


lxfontes (approved):
Our interface map is not the authoritative source of all valid interfaces, but hey it is pretty good!

I wonder if this is where `--force` comes in.
The safest path is to fail  on any validation error, even warnings. Ex:
```
wash app deploy newish-interface.yaml
Warning: Unknown interface "bla"
exit 1
```

vs

```
wash app deploy --force newish-interface.yaml
Warning: Unknown interface "bla"
exit 0
```

brooksmtownsend:
I like that idea @lxfontes , I think we should fail on warnings and then allow forcing past that. My thought is that we should still fail without allowing a `--force` on the actual errors, e.g. an invalid manifest or a link to a component that doesn't exist. Do you think a `--force` should override all of that or still have some failure cases?

lxfontes:
agreed 👍  `--force` should only bypass warnings

thomastaylor312 (approved):

PR #541, opened 2025-01-07
feat(wadm-types): add Hash derive macro for WIT-generated types

LUK3ARK:


LUK3ARK:
Now that I think about it, I think better if we don't attach any extra derives like Hash to the bindgen types, and prefer for all logic to be dealt with the actual types in lib.rs, then anytime it needs to be translated to the wit interface or back use From and Into

reopen if any comment :)

PR #548, opened 2025-01-17
refactor: allow embedding wadm

brooksmtownsend:
## Feature or Problem
This PR reorganizes most of the code from `src/main.rs` to be under the `wadm` crate instead so that Rust developers can embed wadm into their own library. This is primarily useful with the `start_wadm` function, which can be called with a `WadmConfig` struct with all wadm configuration options. If using in a CLI, the `cli` feature should be enabled in the `wadm` crate to enable the clap attributes to work.

## Related Issues
Required for https://github.com/wasmCloud/wasmCloud/issues/1061

## Release Information
wadm-cli 0.20.0
wadm 0.20.0


crates/wadm/src/lib.rs
```diff
@@ -39,3 +73,400 @@ pub const APP_SPEC_ANNOTATION: &str = "wasmcloud.dev/appspec";
...
+///                 }
+///             }
+///         }
+///         _ = tokio::signal::ctrl_c() => {
+///             tracing::info!("Received Ctrl+C, shutting down");
+///             std::process::exit(0);
+///         }
+///     }
+/// ```
+pub async fn start_wadm(config: WadmConfig) -> anyhow::Result<JoinSet<anyhow::Result<()>>> {
```
thomastaylor312, line 120: Nit: Might be worth doing a `use anyhow:Result` up top to make this cleaner

crates/wadm/src/lib.rs
```diff
@@ -39,3 +73,401 @@ pub const APP_SPEC_ANNOTATION: &str = "wasmcloud.dev/appspec";
 pub const SCALER_KEY: &str = "wasmcloud.dev/scaler";
 /// The default link name. In the future, this will likely be pulled in from another crate
 pub const DEFAULT_LINK_NAME: &str = "default";
+pub(crate) const WADM_EVENT_STREAM_NAME: &str = "wadm_events";
```
thomastaylor312, line 76: Super Nit: Should we just make these full `pub` and preface them with `DEFAULT_`? It might be useful to have them around if consuming as a library

crates/wadm/src/nats.rs
```diff
@@ -40,6 +39,16 @@ impl From<StreamPersistence> for StorageType {
...
+        match persistance {
+            "file" => StreamPersistence::File,
+            "memory" => StreamPersistence::Memory,
+            _ => StreamPersistence::File,
+        }
+    }
+}
+
 /// Creates a NATS client from the given options
 pub async fn get_client_and_context(
```
thomastaylor312, line 53: Should this function be `pub` now that it is in a crate we release? I think this is more of an internal function. Probably should also check that for the other functions in this file

thomastaylor312 (changes requested):
Mostly nits but one minor concern around public functions in one of the moved files

thomastaylor312 (approved):

PR #563, opened 2025-01-26
fix(wadm): correct status topic name in WADM

LUK3ARK:
When generating the subscribe wadm topic, it does it incorrectly because it defaults to using the topic prefix of 'wadm.api.{lattice}'

however when subscribing to status we need to use wadm.status

After this fix, this should change:
wadm.status.wadm.api.default.my-app

to:
wadm.status.default.my-app

without affecting any other topic construction


brooksmtownsend (approved):
Yeah this is likely just due to confusion of topic prefix including the lattice or not 🤔 This looks correct though given the fact that the lattice can't include periods. May be better to split the topic prefix from the lattice to avoid confusion in the future

Issue #478, opened 2024-11-07
[BUG] Deadloop for duplicated configuration


peter-jerry-ye:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [X] other / not sure

### Describe the bug

(I'm not sure if it's the problem with wadm or wasmcloud host but) if there are two configurations sharing the same name for the same link, there will be a dead loop.

### Steps to reproduce

Define an application where a link has a configuration with the same name

### Expected behavior

I would expect to see some error / warning, instead of logging insanely `putting configuration` etc.

### Environment

- OS: MacOS
- Shell: zsh
- wasmcloud Version:
```
wash          v0.36.1
├ nats-server v2.10.20
├ wadm        v0.18.0
└ wasmcloud   v1.4.0
```

### Screenshots / Logs / Additional context

![截屏2024-11-04 17 34 27](https://github.com/user-attachments/assets/18eba099-4f51-444f-b699-978eb0c6bfc1)

Issue #401, opened 2024-08-28
[FEATURE] Add shared configs between components / capability


jamesstocktonj1:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

I would like the ability to be able to share configurations between many components and/or capabilities. This can be useful for a range of situations such as having vaguely related services which need to be configured with the same parameters. One such example could be when you have an event producer and an event consumer that need to be configured to communicate on the same topic.

### Describe the solution you'd like

One solution could be a new `config` type as seen in this example config:
```
spec:
  components:
    - name: my-global-config
      type: config
      properties:
        topic: my-topic

    - name: producer
      type: component
      properties:
        image: ghcr.io/username/my-producer:0.1.0
        config:
          - name: my-global-config

    - name: consumer
      type: component
      properties:
        image: ghcr.io/username/my-consumer:0.1.0
        config:
          - name: my-global-config
```
By adding `- name: my-global-config` to the component config this allows both of the components to be configured with the value `topic: my-topic`.

### Describe alternatives you've considered

With this solution we could also allow a kind of "overriding" of config values such as the following:
```
spec:
  components:
    - name: my-global-config
      type: config
      properties:
        foo: bar

    - name: component-a
      type: component
      properties:
        image: ghcr.io/username/my-component-a:0.1.0
        config:
          - name: my-global-config
      # resultant config:
      #   - foo: bar

    - name: component-b
      type: component
      properties:
        image: ghcr.io/username/my-component-b:0.1.0
        config:
          - name: my-global-config
            properties:
              foo: bat
      # resultant config:
      #   - foo: bat
```
This could be a possible solution to avoid overwriting the `my-global-config` on deployment. But this may make things complicated when it comes to subscribing to config changes.

jamesstocktonj1:
Hi Thomas, thanks for moving to the right repo. I hadn't seen that proposal of sharing manifests and I don't see why a `type: config` should behave any differently. That being said could I propose a different solution...

Perhaps a scope value could be assigned to the config to make it crystal clear where it should be visible:
```
- name: my-global-config
    type: config
    scope: global
    properties:
      foo: bar

- name: my-local-config
    type: config
    properties:
[...]

PR #585, opened 2025-02-07
Only check configs for unique names if they declare properties

StuartHarris:
## Feature or Problem
https://github.com/wasmCloud/wadm/pull/516 rejects manifests where multiple link configurations have the same name. This PR augments that functionality by only considering link configurations that declare properties. This is because (as currently implemented) configurations that don't declare properties can reference existing configs that are created either by `wash` or by `wadm` (prefixed by the application name, e.g. where application = "my-app" and config = "my-config", this can be referenced elsewhere in the manifest as "my_app-my_config").

## Related Issues
https://github.com/wasmCloud/wadm/pull/516
https://github.com/wasmCloud/wadm/issues/478
https://github.com/wasmCloud/wadm/issues/401

## Release Information
Manifests that were previously rejected may now be valid. As this widens rather than narrows, I suspect it can be released as a non-breaking minor/patch release.

## Consumer Impact
All manifests that were previously valid should remain valid. 

### Acceptance or Integration
The existing test case for validation of duplicate config names has been adjusted to include properties for the duplicate configs. Additionally, duplicate configs without properties were added to show that they are now allowed through the validation stage.

### Manual Verification
None

### DIFF
--- a/crates/wadm-types/src/validation.rs
+++ b/crates/wadm-types/src/validation.rs
@@ -697,16 +697,21 @@ pub fn validate_component_properties(application: &Manifest) -> Vec<ValidationFa
 /// Validates link configs in a WADM application manifest.
 ///
 /// At present this can check for:
-/// - all config names are unique
+/// - all configs that declare `properties` have unique names
+///   (configs without properties refer to existing configs)
 ///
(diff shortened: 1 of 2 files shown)


brooksmtownsend (approved):
Wonderful work! LGTM 😄

PR #598, opened 2025-03-03
feat(helm): Kubernetes labels

nlamirault:
## Feature or Problem

Support all [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) and additional labels to add to all resources


charts/wadm/templates/_helpers.tpl
```diff
@@ -36,10 +36,15 @@ Common labels
 {{- define "wadm.labels" -}}
 helm.sh/chart: {{ include "wadm.chart" . }}
 {{ include "wadm.selectorLabels" . }}
+app.kubernetes.io/component: manager
```
joonas, line 39: Let’s call this “wadm”

joonas (approved):

PR #670, opened 2025-05-20
feat(types): Added LimitsConfig field to ComponentProperties 

Aditya1404Sal:
for granular per-component resource control

## Feature or Problem

This PR aims to allow the `max_linear_memory` field to be configurable by the wadm.yaml, which will ultimately be used for setting per component resource limits.
## Related Issues

[Memory Limits Per Component](https://github.com/wasmCloud/wasmCloud/issues/4168)

### DIFF
--- a/crates/wadm-types/src/bindings.rs
+++ b/crates/wadm-types/src/bindings.rs
@@ -113,6 +113,7 @@ impl From<ComponentProperties> for wadm::types::ComponentProperties {
             id: properties.id,
             config: properties.config.into_iter().map(|c| c.into()).collect(),
             secrets: properties.secrets.into_iter().map(|c| c.into()).collect(),
+            limits: properties.limits.map(Into::into),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


crates/wadm-types/src/bindings.rs
```diff
@@ -113,6 +113,7 @@ impl From<ComponentProperties> for wadm::types::ComponentProperties {
             id: properties.id,
             config: properties.config.into_iter().map(|c| c.into()).collect(),
             secrets: properties.secrets.into_iter().map(|c| c.into()).collect(),
+            limits: properties.limits.map(Into::into),
```
thomastaylor312, line 116: Heads up that you'll need to add the new field to the wit

Aditya1404Sal:
> Only a spelling nit! LGTM once @thomastaylor312 's comments are addressed.

Got it—I'll have those resolved ASAP 🫡

crates/wadm-types/wit/deps/wadm/types.wit
```diff
@@ -123,6 +123,17 @@ interface types {
         id: option<string>,
         config: list<config-property>,
         secrets: list<secret-property>,
+        limits: list<limits-config>,
```
thomastaylor312, line 126: We should bump the version of the package and add a `since` annotation on these. @brooksmtownsend any preference on whether we do a minor bump or a patch bump? I think with wit, if you add something to a `record` it is a breaking change, right?

